PR #3, opened 2021-08-16
Updgrade dependencies. Generalized library.

adlrocha:
This PR:

- Upgrades `go-ipld-prime` dependencies to use the latest linkSystem.
- Refactors a bit the interface and other methods to add flexibility and allow user-defined logic in subscriber.
- Adds `SubFnHandler` to configure subscriber logic. This handlers can be used to, for instance, select the IPLD selector to use for the exchange.

The reasoning behind this PR is to make it general enough to be able to use `go-legs` as the core for the [indexer ingestion protocol](https://github.com/filecoin-project/indexer-reference-provider/blob/docs/design/docs/design.md).


interface.go
```diff
@@ -20,6 +21,11 @@ type LegPublisher interface {
 
 // LegSubscriber is an interface for watching a published dag.
 type LegSubscriber interface {
+	Subscribe(context.Context, SubFnHandler) error
 	OnChange() (chan cid.Cid, context.CancelFunc)
 	Close(context.Context) error
 }
+
+// SubFnHandler is the function handler to run every time a root is updated
```
willscott, line 29: it seems that this replaces the `OnChange` stream of head updates. Consider removing that?

willscott: can we change `SubFnHandler` to be a function to decide if a `*pubsub.Message` should be considered for a legs update? so something like

```
SetPolicyHandler(PolicyHandler) error
type PolicyHandler func(context.Context, *pubsub.Message) (allowProcessing bool, err error)
```

adlrocha: That actually makes even more sense and it addresses what I was trying to achieve with `SubFnHandler`. Will include it right away. Thanks!

adlrocha: I just realized we may not be able to use `go-legs` for the reference provider if we use a single pubsub channel to broadcast the advertisement of every provider (as long as we don't add additional data in the pubsub message), as indexers we'll need to inspect Advertisements to check if it belongs to a provider they are indexing:
- Provider publishes Advertisement CID in common PubSub channel.
- Indexer sees CID for a new advertisement and starts a data transfer to fetch just the advertisement.
- Indexer checks if advertisement is from a provider it is indexing.
- If yes, indexer fetches the full DAG recursively and stops in the last advertisement seen for that provider.

I don't think we can embed this logic in a `SetPolicyHandler` or a `SubFnHandler`, we'll need to implement an adaptation of `go-legs` for the provider. What do you think?

willscott: won't the indexer know if it's worth pulling based on the libp2p identity of the provider that it connects to in step 2?

adlrocha: Oh wait! I think you are right. Do we have a way to check the peer that originally published a pubsub message to a topic from a`pubsub.Message`? I thought the `msg.From` of `*pubsub.Message` specified who you were receiving the message from, not the one that generated it. If we have a way to know from the message who published then forget what I said 🤦‍♂️ (and it seems so because I see there's a `GetFrom()` and a `ReceivedFrom()` in the pubsub code).

willscott: https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L218 ?

adlrocha:
@willscott, added `PolicyHandler` and a few things more. I did some tests and I think this is looking good for what we need in the indexer reference implementation. Thanks for the feedback.

interface.go
```diff
@@ -20,6 +23,23 @@ type LegPublisher interface {
 
 // LegSubscriber is an interface for watching a published dag.
 type LegSubscriber interface {
+	Subscribe(ctx context.Context, selector ipld.Node, policy PolicyHandler) error
```
willscott, line 26: I'm not convinced there should be a `Subscribe` method on the interface.
* subscribing to the pubsub topic and configuration there can live outside of the interface in the specific implementation constructor.
* perhaps there's a hook for `NoteUpdatedSyncState(head cid.Cid)` to short-circuit some of the subsequent announcement fetch that has been loaded through other means.

I think the main thing is that i think we should make construction and ownership over the selector be something that happens within this library, rather than something set by the client using it.
Is there something in the customization of that selector you think the client making use of legs will do better than the library itself?

adlrocha: Makes sense. I don't have a strong opinion on this, so happy to do otherwise. 

My reasoning behind this was that users may want to use legs in a general way for different type of exchanges when an update for a DAG is seen. In our case we explore recursively and use a dynamic stop condition, but other subscribers may want to subscribe with a selector that exchanges all from scratch, or which fetches the node for the latest update.

If I understood correctly, your suggestion then is to set the `ExploreRecursive` selector as the one used for legs subscribers by design, and add additional config through hooks or when instantiating it, removing `Subscribe` from the interface, right?

willscott: yeah. I think lets wait until we have clients wanting other behavior to figure out what the right interface is for that. i'd guess in general it'll be part of the constructor rather than on the interface, since it's unlikely to change dynamically within a session

willscott (approved):

PR #4, opened 2021-08-18
Add partial sync support.

adlrocha:
This PR:
- Fixes the subscriber selector to `ExploreRecursiveWithStop` in order to be able to perform partial sync.
- Adds `latestSync` in subscribers so when there is a new update they don't need to sync the DAG from scratch if they've already seen some part of it.
- Adds a bunch of tests to see if everything works.

### DIFF
--- a/sync_test.go
+++ b/sync_test.go
@@ -1,0 +1,207 @@
+package legs
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"io/ioutil"
+	"testing"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	leveldb "github.com/ipfs/go-ds-leveldb"
+	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/fluent"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	basicnode "github.com/ipld/go-ipld-prime/node/basic"
+	"github.com/libp2p/go-libp2p"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/multiformats/go-multicodec"
+)
+
+func mkTestHost() host.Host {
+	h, _ := libp2p.New(context.Background(), libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/0"))
+	return h
+}
+
+func mkLinkSystem(ds datastore.Batching) ipld.LinkSystem {
+	lsys := cidlink.DefaultLinkSystem()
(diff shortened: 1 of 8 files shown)


sync_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"bytes"
+	"context"
+	"io"
+	"io/ioutil"
+	"testing"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	leveldb "github.com/ipfs/go-ds-leveldb"
```
willscott, line 13: can we use a mem db since this is only for tests and lets us avoid otherwise pulling in the dependency?

willscott (approved):

sync_test.go
adlrocha: Do we have any memory datastore that is "concurrent-friendly"? (I don't know of any). I had to use levelDB parallel read/writes in the MapDatastore for one of the tests (is the only one I knew).

willscott: That might be true.

is it easy to switch this file to be in a `legs_test` package?

sync_test.go
```diff
@@ -0,0 +1,207 @@
...
+	dstore, err := leveldb.NewDatastore(tmpDir, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return dstore
+}
+func TestLatestSync(t *testing.T) {
+	// NOTE: In order to aviod parallel read/write (which makes the test flaky), we need
+	// to use a concurrency-friendly datastore like levelDB
+	srcStore := mkLevelDs(t)
```
willscott, line 119: make sure to defer a cleanup of files that might be written?

sync_test.go
adlrocha: I don't think we can without exposing `latestSync` through the `Subscriber` interface. One thing I can probably do is to create a new `MapDatastore` in each step of the test and load the part of the dag already synced to avoid having to use LevelDB. It is a bit more of work, but I'll look into this.

willscott: It's not the end of the world, just hoping to not pull it into the dependency tree unnecessarily

adlrocha: No, you are right. It's not a great idea to pull LevelDB just for testing when users will actually end-up hooking their preferred datastore when using the library. I'll look into it tomorrow.

adlrocha:
@willscott, it took me more than I'd wish for, but now we can do consistent testing without pulling additional datastores. I am relatively happy with the tests now. Let me know if you see anything missing. Thanks!

willscott (approved):

PR #10, opened 2021-08-24
Do not delete the system TMPDIR

gammazero:
This change makes publish and subscribe create their own temp dir and delete it on Close, instead of using the system temp dir and then trying to delete that (causing all of a user's files in that directory to get deleted).

### DIFF
--- a/publish.go
+++ b/publish.go
@@ -26,6 +26,9 @@ type legPublisher struct {
 }
 
 // NewPublisher creates a new legs publisher
+//
+// TODO: Add a parameter or config to set the directory that the publisher's
+// tmpDir is created in
 func NewPublisher(ctx context.Context, dataStore datastore.Batching, host host.Host, topic string, lsys ipld.LinkSystem) (LegPublisher, error) {
 
 	t, err := makePubsub(ctx, host, topic)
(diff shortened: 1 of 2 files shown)


publish.go
```diff
@@ -37,7 +38,10 @@ func NewPublisher(ctx context.Context, dataStore datastore.Batching, host host.H
 	tp := gstransport.NewTransport(host.ID(), gs)
 	dtNet := dtnetwork.NewFromLibp2pHost(host)
 
-	tmpDir := os.TempDir()
+	tmpDir, err := ioutil.TempDir("", "golegs-pub")
```
masih, line 41: Would it make sense to parameterise the tmpDir and fall back onto some default if unspecified?

The rationale is that during testing a temp dir could be set that's created via `testing.T`

masih (approved):

publish.go
gammazero: Yes.  Is it OK to make a breaking interface change (new parameter to New...) or should there be a separate New... that takes an additional parameter, or change New... to take config struct or variadic options?

masih: mm... good question. Since parameterisation of `tmpDir`is unrelated to the fix I'd say let's leave a `TODO` and defer to a separate PR

gammazero: Done.  Added TODO.

PR #12, opened 2021-08-29
Single data transfer to support multiple subscribers

adlrocha:
* Add `LegTransport` as common transfer layer for publisher and subscribers.
* Refactor publish and subscribers to use a `LegTransport`.
* Fix syncing problem when multiple subscribers listen to events from the common data transfer.


transport.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	gsnet := gsnet.NewFromLibp2pHost(host)
+	gs := gsimpl.New(context.Background(), gsnet, lsys)
+	tp := gstransport.NewTransport(host.ID(), gs)
+	dtNet := dtnetwork.NewFromLibp2pHost(host)
+
+	tmpDir, err := ioutil.TempDir("", tmpPath)
```
willscott, line 54: is this for testing, or do we always call this method to setup a transport?
Can we let the client directly specify (and clean up) tmpDir?

transport.go
```diff
@@ -0,0 +1,119 @@
...
+
+// LegTransport wraps all the assets to set-up the data transfer.
+type LegTransport struct {
+	tmpDir string
+	t      dt.Manager
+	Gs     graphsync.GraphExchange
+	topic  *pubsub.Topic
+
+	lk   sync.Mutex
+	refc int
```
willscott, line 38: consider updates to `refc` via [atomic](https://pkg.go.dev/sync/atomic) rather than protecting it with a mutex?

transport.go
```diff
@@ -0,0 +1,119 @@
...
+	defer lt.lk.Unlock()
+	lt.refc++
+}
+
+func (lt *LegTransport) Close(ctx context.Context) error {
+	lt.lk.Lock()
+	defer lt.lk.Unlock()
+	lt.refc--
+
+	if lt.refc == 0 {
```
willscott, line 90: this is still racy - if one transfer closes and another opens, you don't know if you've actually cleaned up or not depending on the ordering

transport.go
adlrocha: This is called for every transport, not only for testing. As this tmpDir is only used for the data channel, I wanted to make it as transparent as possible for the client (I don't think it makes sense keeping this directory or sharing it once the data channel is closed). 

Your suggestion is, instead of creating a new dir in the path specified, to let clients provide the actual tmp path, create it if doesn't exist, and let them handle it from there on (so not removing it in `Close()`?

adlrocha: Follow-up question, do you think that instead of closing the transport when it has no more subscribers or publishers should we let the client close when it doesn't need it anymore? (maybe throwing an error if there is any active subscriber/publisher). Someone may want to reuse a transport again even after removing all subscribers from the transport, right? (an this would also simplify the `refc` management.

willscott: I started trying to figure out why we need this and i _think_ it's deprecated and un-used such that it doesn't matter at this point.

The argument is for the cidlist directory, and at one point the list of requested cids was being written as a file in a given directory. That code appears deprecated, and at least at the head of graphsync code, seems to be replaced in the 'channel' code with `cidset` which will use a namespace within the provided datastore to track cids, rather than a flat file (which seems better).

Can we confirm that's true? if so, then we can ignore this with just a random name, and shouldn't expose it upwards

willscott (approved):

PR #14, opened 2021-09-01
Support proactive sync from subscribers

### LINKED ISSUE #5: Update README examples
Once the library is stabilized we need to update the examples in the README (with the latest changes they are currently broken).

### LINKED ISSUE #13: Use t.Cleanup instead of defer in tests
Example: https://github.com/willscott/go-legs/blob/3139ce32c8a352a50baf739dfc9083ead4a3ff64/legs_test.go#L117

adlrocha:
This PR:
 * Adds `Sync` to subscribers so they can proactively sync with a publisher without having to wait for an update to the pubsub channel.
 * Fix #13 
 * Fix #5 

- Discussion:
`Sync` updates the latest head seen in the subscriber. This means that if a subscribers chooses to sync to a head that is older than the latest sync for a publisher it has already seen, it may need to re-fetch parts of the DAG it has already stored in the next update. My take is that this should be well-documented and handled by the library using `go-legs` and not `go-legs` itself. In any case, wanted to flag it here in case there are other opinions/ideas on the matter.


subscribe.go
```diff
@@ -230,3 +230,38 @@ func (ls *legSubscriber) Close() error {
...
+		if event.Code == dt.FinishTransfer {
+			if c == channelState.BaseCID() {
+				out <- channelState.BaseCID()
+				log.Debugw("Exchange finished, updating latest to  %s", channelState.BaseCID())
+				// Update latest sync to the head we are proactively syncing to.
+				// Beware! This means that if we sync to a CID which is before
+				// the latest one it may require to resend content from the chain that
+				// we already have. This should be handled by users.
+				ls.latestSync = cidlink.Link{Cid: channelState.BaseCID()}
+				ls.syncmtx.Unlock()
```
willscott, line 263: In cases where the sync is canceled, and the channel is unsubscribed, the mutex will never be unlocked. Is this expected?

adlrocha: Wow! You are right. I didn't think about that case. Fixed and added a test to check this. Thanks.

mutex.go
```diff
@@ -0,0 +1,20 @@
...
+	m.lk.Lock()
+	m.isLocked = true
+}
+
+func (m *syncMutex) unlock() {
+	if m.isLocked {
+		m.lk.Unlock()
+		m.isLocked = false
+	}
+}
```
willscott, line 20: why would we be calling Unlock more than once on a Lock? The panic on doing that is meant to push you to structure your code to not attempt it, rather than make the call to Unlock optional.

And noting that that this construction introduces a potential issue:
in a state where the lock is 'locked', and 2 goroutines, 1 calling lock and one unlock:
* [lock] blocks on the 'Lock()' call
* [unlock] calls 'Unlock'
* [lock] finishes 'Lock()' call. sets `isLocked` to true
* [unlock] sets `isLocked` to false 

you now have a locked mutex, but `isLocked` set to false.

It seems like the ultimate thing you're trying to do is to make sure there's only 1 transfer going at once here, right?
consider maybe a `sync.Once` that's called on error/completion as a way to set the state back to a world where the next action is allowed?

adlrocha: I wasn't aware of the existence of `sync.Once`, thank you for the note. The problem with the double unlock was that by exposing `cancelFunc` we take the risk of someone calling `cancelFunc()` and thus unlocking the lock that has been already unlocked when the transfer finished (the tests consider that case, so in the future we should see it blow up if it happens). I think I fixed it now.

And you are right, my mutex construction had parallelization issues I didn't consider, I would have needed to improve/test it to prevent the kind of problems you mentioned.

subscribe.go
```diff
@@ -230,3 +230,52 @@ func (ls *legSubscriber) Close() error {
...
+	ls.syncmtx.Lock()
+	out := make(chan cid.Cid)
+	var ulOnce sync.Once
+
+	v := Voucher{&c}
+	unsub := ls.transfer.t.SubscribeToEvents(ls.onSyncEvent(c, out, &ulOnce))
+	_, err := ls.transfer.t.OpenPullDataChannel(ctx, p, &v, c, legSelector(nil))
+	if err != nil {
+		log.Errorf("Error in data channel for sync: %v", err)
+		ls.syncmtx.Unlock()
```
willscott, line 250: should this also be the `unlockOnce` call?

subscribe.go
```diff
@@ -230,3 +230,52 @@ func (ls *legSubscriber) Close() error {
 	ls.cancel()
 	return nil
 }
+
+func (ls *legSubscriber) unlockOnce(ulOnce *sync.Once) {
+	ulOnce.Do(func() {
```
willscott, line 235: can this directly be `ulOnce.Do(ls.syncmtx.Unlock)` ?

subscribe.go
adlrocha: It'd make no difference because we are not returning `CancelFunc` yet, so there is no option for someone to cancel and trying to unlock twice.

willscott (approved):

PR #18, opened 2021-09-23
simplify the use of atomics over refc

mvdan:
(see commit message)


mvdan:
Spotted while I was attempting to pull "wait for enough peers" from pubsub to indexer-reference-provider. Seems worthwhile.

willscott (approved):

PR #19, opened 2021-10-01
Decouple publishing from instantiating data transfer

hannahhoward:
# Goals

In the reference provider, we need to serve go-legs while also serving retrievals (this will be needed in filecoin integration as well). This means we can't instantiate go-data-transfer inside the LegsTransport. And we don't want legsTransport.Close to shut the whole thing down either.

# Implementation

I'm not sure if this is the right approach -- ideally, what I'd like to do is get rid of LegsTransport altogether -- I'm not sure I understand the value of this abstraction, especially since it's main method is not used.

BUT we have code expecting go-legs to behave the way it does so I chose an initial path that avoids breaking anything. Open to other suggestions.

### DIFF
--- a/publish.go
+++ b/publish.go
@@ -2,31 +2,36 @@ package legs
 
 import (
 	"context"
-	"sync/atomic"
 
 	"github.com/ipfs/go-cid"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
 )
 
 type legPublisher struct {
-	transfer *LegTransport
+	r     Reference
+	topic *pubsub.Topic
 }
 
 // NewPublisher creates a new legs publisher
 func NewPublisher(ctx context.Context, dt *LegTransport) (LegPublisher, error) {
+	return NewPublisherFromDeps(ctx, dt, dt.topic)
+}
 
-	// Track how many publishers are using this transport
-	dt.addRefc()
-
+// NewPublisherFromDeps instantiates go-legs from direct dependencies
+func NewPublisherFromDeps(ctx context.Context, r Reference, topic *pubsub.Topic) (LegPublisher, error) {
+	r.AddReference()
 	return &legPublisher{
-		transfer: dt}, nil
+		r:     r,
+		topic: topic,
+	}, nil
 }
 
 func (lp *legPublisher) UpdateRoot(ctx context.Context, c cid.Cid) error {
 	log.Debugf("Published CID in pubsub channel: %s", c)
-	return lp.transfer.topic.Publish(ctx, c.Bytes())
+	return lp.topic.Publish(ctx, c.Bytes())
 }
 
 func (lp *legPublisher) Close() error {
-	atomic.AddInt32(&lp.transfer.refc, -1)
+	lp.r.RemoveReference()
 	return nil
 }
--- a/subscribe.go
+++ b/subscribe.go
@@ -3,7 +3,6 @@ package legs
 import (
 	"context"
 	"sync"
-	"sync/atomic"
 
 	dt "github.com/filecoin-project/go-data-transfer"
 	"github.com/ipfs/go-cid"
@@ -74,7 +73,7 @@ func newSubscriber(ctx context.Context, dt *LegTransport, policy PolicyHandler) 
 		return nil, err
 	}
 	// Add refC to track how many subscribers are using the transport.
-	dt.addRefc()
+	dt.AddReference()
 
 	return ls, nil
 }
@@ -229,7 +228,7 @@ func (ls *legSubscriber) OnChange() (chan cid.Cid, context.CancelFunc) {
 }
 
 func (ls *legSubscriber) Close() error {
-	atomic.AddInt32(&ls.transfer.refc, -1)
+	ls.transfer.RemoveReference()
 	ls.cancel()
 	return nil
 }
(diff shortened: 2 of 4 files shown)


hannahhoward:
another alternative would be to have an interface that matches whatever is needed:

type LegTransport interface {
   Reference
   DataTransfer() dt.Manager
   Topic() *pubsub.Topic
}

That could enable the MakeLegTransport method while allowing someone else to implement another version.

willscott:
We're probably partially running into edges that are beyond the initial complexity legs was aiming to handle, which was a very simple interface to let a publisher and set of subscribers keep an IPLD dag in sync.

At the top level, it seems like the thing we'd like to do is be able to pass in the existing go-data-transfer along with the host if desired when making the legs handler for that.

The initial design didn't expose a 'legs transport' at all, and i'm happy with that evolving to be hidden back in to the top level goals a user would have of 'make a publisher' and 'make a subscriber' against a topic and a host. The machinery/complexity exposed in the transport were I believe a response to trying to handle multiplexing on the publisher side.

multisubscribe.go
```diff
@@ -0,0 +1,124 @@
...
+	ctx    context.Context
+	tmpDir string
+	t      dt.Manager
+	gs     graphsync.GraphExchange
+	topic  *pubsub.Topic
+	refc   int32
+}
+
+// NewMultiSubscriber sets up a new instance of a multi subscriber
+func NewMultiSubscriber(ctx context.Context, host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topic string) (LegMultiSubscriber, error) {
```
willscott, line 44: this is for use in the indexer, presumably. is this an interface that storetheindex will be able to integrate? since that's the downstream that we should probably update with this interface change

publish.go
```diff
@@ -2,31 +2,56 @@ package legs
...
+		return nil, err
+	}
+	return &legPublisher{ss.t, ss.onClose}, nil
+}
 
-	return &legPublisher{
-		transfer: dt}, nil
+// NewPublisherFromExisting instantiates go-legs publishing on an existing
+// data transfer instance
+func NewPublisherFromExisting(ctx context.Context,
```
willscott, line 34: the thought is that this is what we would expect a lotus to make use of, right?

subscribe.go
```diff
@@ -34,19 +37,33 @@ type legSubscriber struct {
 
 // NewSubscriber creates a new leg subscriber listening to a specific pubsub topic
 // with a specific policyHandle to determine when to perform exchanges
-func NewSubscriber(ctx context.Context, dt *LegTransport,
-	policy PolicyHandler) (LegSubscriber, error) {
-	return newSubscriber(ctx, dt, policy)
+func NewSubscriber(ctx context.Context,
```
willscott, line 40: should we support a `NewSubscriberFromExisting`?

utils.go
```diff
@@ -0,0 +1,155 @@
+package legs
```
willscott, line 1: there are a few different unrelated things in this file. consider splitting it up:
* simple setup
* the voucher stuff

transport.go
```diff
@@ -1,166 +0,0 @@
...
-
-func (lt *LegTransport) addRefc() {
-	atomic.AddInt32(&lt.refc, 1)
-}
-
-// Fetch starts a new transfer with a peer for certain CID using the selector passed as argument.
-//
-// sel determines the selector to use for the exchange
-// Optionally, we can run an onFinish callback when the exchange has finished.
-func (lt *LegTransport) Fetch(ctx context.Context, p peer.ID, c cid.Cid, sel ipld.Node, onFinish func()) (chan cid.Cid, context.CancelFunc, error) {
```
adlrocha, line 101: I think the `MultiSubscriber` approach is a great idea. The purpose of the `LegTransport` was just to offer an abstraction of pubsub+data-transfer making them transparent for the user. My main concern is that  `LegTransport` included a `Fetch` function to proactively fetch a graph from any peer without requiring an advertisment. This is something we'll eventually need to implement step-by-step syncing in storetheindex. We can probably include this function as part of the multisubscriber interface.

multisubscribe.go
adlrocha: Yes, we'll need to propagate these changes to storetheindex. I think they may be quite straightfoward once we implement `Fetch` in MultiSubscriber.

adlrocha (changes requested):

publish.go
hannahhoward: yea, and honestly the reference provider would do well to do it also

subscribe.go
hannahhoward: I don't think until we have a use case

transport.go
hannahhoward: @adlrocha why would this function exist here? If you have access to the data transfer (which you do on the current multi-subscriber interface, why not just implement it there. I don't understand what is "go-legs" specific about this -- it appears to be simply a data transfer call

willscott: i think from the provider perspective the right thing to do will be to limit to the two actually used selector forms: "give me the chain of advertisements back to some cid" and "give me the cids of a specific advertisement.

adlrocha: That's a great point, @hannahhoward. The idea was to try an abstract the developer in `go-legs` from vouchers, and the underlying complexity of `go-data-transfer`. I actually like @willscott's idea, the same way that we are currently exposing `LegSelector` we can limit the selectors that can be used within `go-legs` (in the end we'll only potentially use two of them for the subscriber).

PR #23, opened 2021-10-15
http-transport legs

willscott:
Conceptually, provide the same publisher and subscriber interface as legs, but with subscriber pulling from static files that the publisher makes available in an HTTP directory.

* does not handle alerts of updated head in-band. a second notification protocol would be needed, as this will either poll, or use manual calls to`Sync` to fetch.

### DIFF
--- a/http/subscribe.go
+++ b/http/subscribe.go
@@ -1,0 +1,164 @@
+package http
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"path"
+	"sync"
+
+	"github.com/filecoin-project/go-legs"
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	basicnode "github.com/ipld/go-ipld-prime/node/basic"
+	peer "github.com/libp2p/go-libp2p-peer"
+	"github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+	"github.com/multiformats/go-multicodec"
+)
+
+// NewHTTPSubscriber creates a legs subcriber that provides subscriptions
+// from publishers identified by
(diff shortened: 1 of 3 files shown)


http/subscribe.go
```diff
@@ -0,0 +1,305 @@
...
+			LinkTargetNodePrototypeChooser: basicnode.Chooser,
+		},
+		Path: datamodel.NewPath([]datamodel.PathSegment{}),
+		LastBlock: struct {
+			Path datamodel.Path
+			Link datamodel.Link
+		}{
+			Path: ipld.Path{},
+			Link: nil,
+		},
```
gammazero, line 247: Since these are the zero values, the do not need to be explicitly assigned:
```go
	LastBlock: struct {
		Path datamodel.Path
		Link datamodel.Link
	}{},
```

http/subscribe.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	// get the direct node.
+	rootNode, err := getMissingLs.Load(ipld.LinkContext{}, cidlink.Link{Cid: root}, basicnode.Prototype.Any)
+	if err != nil {
+		return err
+	}
+	return progress.WalkAdv(rootNode, sel, func(p traversal.Progress, n datamodel.Node, vr traversal.VisitReason) error {
```
gammazero, line 254: Please comment why `WalkAdv` is used instead of `WalkMatching`, or if that even makes a difference here.

subscribe.go
```diff
@@ -75,16 +79,16 @@ func NewSubscriberPartiallySynced(
...
-		topic:   topic,
-		onClose: onClose,
-		updates: make(chan cid.Cid, 5),
-		subs:    make([]chan cid.Cid, 0),
-		cancel:  nil,
-		policy:  policy,
+		dt:              dt,
+		topic:           topic,
+		onClose:         onClose,
+		updates:         make(chan cid.Cid, 5),
```
gammazero, line 87: Please specify why a channel of size 5 is used.  Or, change it to either 0 or 1.

PR #26, opened 2021-11-08
support pubsub.PubOpts in UpdateRoot

mvdan:
(see commit message)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/ipld/go-ipld-prime v0.12.3
 	github.com/libp2p/go-libp2p v0.15.0
 	github.com/libp2p/go-libp2p-core v0.9.0
-	github.com/libp2p/go-libp2p-pubsub v0.5.4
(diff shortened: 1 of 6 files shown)


willscott (approved):

mvdan:
Oops, hadn't noticed CI was red until just now. The build is now fixed.

Unfortunately, TestLatestSyncSuccess seems to now hang - presumably because EnoughPeers(1) never becomes true. I'm not really sure how that is happening. I'm trying to debug a little.

masih (approved):
My vote is to merge this if it makes the tests more stable despite the hanging test

mvdan:
Well, I certainly can't break master in a reproducible way :)

gammazero:
As documented in [`WithReadiness`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#WithReadiness), I don't think this is going to work without providing a [`WithDiscover`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#WithDiscovery) to give the pubsub instance a discovery mechanism.

PR #30, opened 2021-11-12
Fix race condition in sync tests

masih:
Fix a race condition where gorutines compete to read the value of latest
synced link by introducing a read-lock.

Update test timeouts slightly to accommodate CI run tests.

Refactor tests to reduce duplicate code.

Relates to:
 - https://github.com/filecoin-project/go-legs/pull/29

### DIFF
--- a/subscribe.go
+++ b/subscribe.go
@@ -32,6 +32,7 @@ type legSubscriber struct {
 	policy          PolicyHandler
 	defaultSelector ipld.Node
 
+	// syncmtx synchronizes read/write for latestSync and syncing
 	syncmtx    sync.Mutex
 	latestSync ipld.Link
 	syncing    cid.Cid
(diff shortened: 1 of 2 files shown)


subscribe.go
```diff
@@ -32,7 +32,7 @@ type legSubscriber struct {
 	policy          PolicyHandler
 	defaultSelector ipld.Node
 
-	syncmtx    sync.Mutex
+	syncmtx    sync.RWMutex
```
mvdan, line 35: please document what this mutex locks; I imagine it's the two fields below, but I'm not 100% sure.

subscribe.go
```diff
@@ -305,3 +305,11 @@ func (ls *legSubscriber) onSyncEvent(c cid.Cid, out chan cid.Cid, ulOnce *sync.O
 		}
 	}
 }
+
+// getLatestSync gets the latest synced link.
+// This function is safe to call from multiple goruties and exposed for testing purposes only.
```
mvdan, line 310:
```suggestion
// This function is safe to call from multiple goroutines and exposed for testing purposes only.
```

Unless you're trying to coin a new term :P

mvdan (approved):

subscribe.go
mvdan: By the way, a RWMutex has worse performance than a Mutex if we mostly just grab write locks - because it is a more complex mutex. So I would say it's best to leave this as a sync.Mutex, as only the tests need a read-only lock. Otherwise you might be making legSubscriber slightly slower in terms of lock contention.

mvdan: In other words, a RWMutex is only worthwhile if the majority of your locking operations are read-only. If the majority are read-write, then a Mutex will simply perform better.

masih: sounds good thanks Daniel

PR #31, opened 2021-11-16
Assert peers eventual connectivity in tests

masih:
Assert that peers participating in tests eventually appear in the list
of connected peers for both sides, i.e. publisher and subscriber.

Refactor tests to assert error inplace when constructing components
needed for testing.


masih:
Disclaimer: This PR is for me to investigate the root cause of flaky tests in go-legs, and help myself understand what's happening. 
The eventual assertion introduced in this PR is effectively not needed now that there is a dedicated option for it upstream, where pubsub topic can be instructed to wait a minimum number of peers, introduced in #26.

PR #32, opened 2021-11-17
Add ability to query latest root cid

MarcoPolo:
## Sumary

Adds the ability to query a peer publisher for their current root cid. Which enables [`LegSubscriber.Sync`](https://github.com/filecoin-project/go-legs/blob/main/interface.go#L41) to accept a `cid.Undef` and have it request the current root from the publisher.

## Context

`storetheindex` implemented a version of this to [get the latest advertised cid](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/ingest.go#L140). It's implemented in [`providerclient`](https://github.com/filecoin-project/storetheindex/blob/main/providerclient/interface.go) and in [`index-provider`](https://github.com/filecoin-project/index-provider/tree/main/server/provider/libp2p).

This functionality seems generally useful to any user of go-legs, so there's a desire to refactor the above logic and put it into go-legs.

## Proposed Solution

Very similar to how the `go-legs` [http publisher](https://github.com/filecoin-project/go-legs/tree/main/http) works, this leverages an http server and client backed by a libp2p stream with [go-libp2p-gostream](https://github.com/libp2p/go-libp2p-gostream). The HTTP + gostream solution was chosen to leverage Go's `net` stack and piggieback of the HTTP protocol. While HTTP does add a bit of overhead to the messages being passed, I think it's a worthy tradeoff compared to the work involved in doing this from scratch (even with pb messages). For comparison look at all the helpers and utils involved in just the client side of the [`providerclient` implementation](https://github.com/filecoin-project/storetheindex/blob/main/providerclient/libp2p/client.go#L10).

## Next steps

(in a separate PR) delete all the provider client/server code.

I'm not sure how often this is called, but if it's called enough, storing the `AdvClient` in an LRU cache might be nice.

r? @willscott

### DIFF
--- a/p2p/protocol/adv/adv.go
+++ b/p2p/protocol/adv/adv.go
@@ -66,7 +66,7 @@ func NewAdvClient(ctx context.Context, host host.Host, topic string, peer peer.I
 	}, err
 }
 
-func (c *AdvClient) QueryLatestCid(ctx context.Context) (cid.Cid, error) {
+func (c *AdvClient) QueryRootCid(ctx context.Context) (cid.Cid, error) {
 	// The httpclient expects there to be a host here. `.invalid` is a reserved
 	// TLD for this purpose. See
 	// https://datatracker.ietf.org/doc/html/rfc2606#section-2
--- a/p2p/protocol/adv/adv.go
+++ b/p2p/protocol/adv/adv.go
@@ -1,0 +1,122 @@
+package adv
+
+import (
+	"context"
+	"encoding/json"
+	"net"
+	"net/http"
+	"path"
+	"sync"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/libp2p/go-libp2p-core/host"
+	peer "github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+	gostream "github.com/libp2p/go-libp2p-gostream"
+)
+
+var log = logging.Logger("go-legs-adv")
+
+func deriveProtocolID(topic string) protocol.ID {
+	return protocol.ID("/legs/adv/" + topic + "/0.0.1")
+}
+
+func (p *AdvPublisher) Serve(ctx context.Context, host host.Host, topic string) error {
+	l, err := gostream.Listen(host, deriveProtocolID(topic))
+	if err != nil {
+		return err
+	}
+
+	p.server = &http.Server{Handler: http.Handler(p)}
+	errChan := make(chan error)
+	go func() {
+		errChan <- p.server.Serve(l)
+	}()
+
+	select {
+	case err := <-errChan:
+		return err
+	case <-ctx.Done():
+		return p.server.Close()
+	}
+}
+
+func newHttpClient(ctx context.Context, host host.Host, topic string, peer peer.ID) (http.Client, error) {
+	transport := http.Transport{
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			return gostream.Dial(ctx, host, peer, deriveProtocolID(topic))
+		},
+	}
+
+	return http.Client{
+		Transport: &transport,
+	}, nil
+}
+
+type AdvClient struct {
+	client *http.Client
+}
+
+func NewAdvClient(ctx context.Context, host host.Host, topic string, peer peer.ID) (*AdvClient, error) {
+	client, err := newHttpClient(ctx, host, topic, peer)
+	return &AdvClient{
+		client: &client,
+	}, err
+}
+
+func (c *AdvClient) QueryLatestCid(ctx context.Context) (cid.Cid, error) {
+	// The httpclient expects there to be a host here. `.invalid` is a reserved
+	// TLD for this purpose. See
+	// https://datatracker.ietf.org/doc/html/rfc2606#section-2
+	resp, err := c.client.Get("http://unused.invalid/head")
+	if err != nil {
+		return cid.Undef, err
+	}
+	defer resp.Body.Close()
+
+	var cidStr string
+	err = json.NewDecoder(resp.Body).Decode(&cidStr)
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	return cid.Decode(cidStr)
+}
+
+type AdvPublisher struct {
+	rl     sync.RWMutex
+	root   cid.Cid
+	server *http.Server
+}
+
+func (p *AdvPublisher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ask := path.Base(r.URL.Path)
+	if ask == "head" {
+		p.rl.RLock()
+		defer p.rl.RUnlock()
+		out, err := json.Marshal(p.root.String())
+		if err != nil {
+			w.WriteHeader(500)
+			log.Infow("failed to serve root", "err", err)
+		} else {
+			_, _ = w.Write(out)
+		}
+	} else {
+		w.WriteHeader(404)
+	}
+}
+
+func (p *AdvPublisher) UpdateRoot(_ context.Context, c cid.Cid) error {
+	p.rl.Lock()
+	defer p.rl.Unlock()
+	p.root = c
+	return nil
+}
+
+func (p *AdvPublisher) Close() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	return p.server.Shutdown(ctx)
+}
--- a/p2p/protocol/adv/adv_test.go
+++ b/p2p/protocol/adv/adv_test.go
@@ -12,16 +12,15 @@ import (
 	_ "github.com/ipld/go-ipld-prime/codec/dagjson"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	"github.com/ipld/go-ipld-prime/node/basicnode"
-	blhost "github.com/libp2p/go-libp2p-blankhost"
-	swarmt "github.com/libp2p/go-libp2p-swarm/testing"
+	"github.com/libp2p/go-libp2p"
 )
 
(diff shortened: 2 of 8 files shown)


p2p/protocol/adv/adv_test.go
```diff
@@ -0,0 +1,54 @@
...
+	blhost "github.com/libp2p/go-libp2p-blankhost"
+	swarmt "github.com/libp2p/go-libp2p-swarm/testing"
+)
+
+func TestFetchLatestAdv(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	publisher := blhost.NewBlankHost(swarmt.GenSwarm(t, ctx))
+	client := blhost.NewBlankHost(swarmt.GenSwarm(t, ctx))
```
willscott, line 24: Can you use `libp2p.New` and avoid the additional swarm/blankhost dependencies like https://github.com/filecoin-project/go-legs/blob/main/legs_test.go#L29 ?

p2p/protocol/adv/adv.go
```diff
@@ -0,0 +1,122 @@
+package adv
```
willscott, line 1: I'm not convinced by this package name.

maybe the protocol would make sense as 'head'? that's what this endpoint is called in the HTTP implementation - may be nice to have the two harmonized.

p2p/protocol/adv/adv.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func NewAdvClient(ctx context.Context, host host.Host, topic string, peer peer.ID) (*AdvClient, error) {
+	client, err := newHttpClient(ctx, host, topic, peer)
+	return &AdvClient{
+		client: &client,
+	}, err
+}
+
+func (c *AdvClient) QueryRootCid(ctx context.Context) (cid.Cid, error) {
```
willscott, line 69: Since we aren't keeping the http client as a statefull object, can we collapse this all to just 1 function?

You can create an http.Client with the custom transport and save some extra indirection.

p2p/protocol/adv/adv_test.go
MarcoPolo: sure! Does this dependency show up for users of the library? (even if it's just used in the test?)

p2p/protocol/adv/adv.go
MarcoPolo: agreed about name, I actually meant to bring that up in the PR description.

`head` seems fine. It's pretty specific, but this thing only does that, so works for me.

p2p/protocol/adv/adv_test.go
willscott: go.mod is getting better at this, but it's still not great and test dependencies do often still make it in to the dependency tree.

publish.go
```diff
@@ -26,7 +28,11 @@ func NewPublisher(ctx context.Context,
 	if err != nil {
 		return nil, err
 	}
-	return &legPublisher{ss.t, ss.onClose}, nil
+
+	headPublisher := &head.Publisher{}
+	go headPublisher.Serve(ctx, host, topic)
```
gammazero, line 33: Passing the ctx into `headPublisher.Serve` does not accomplish anything.  It only makes the `Serve` function stop waiting on its internal goroutine, but does not stop that internal goroutine.  See earlier comment on `Server`.  Also, if then error returned is always ignored, then why return it. Maybe log it, or just don't return it if is not intended to be used.

p2p/protocol/head/head_test.go
```diff
@@ -0,0 +1,49 @@
...
+	client.Peerstore().AddAddrs(publisher.ID(), publisher.Addrs(), time.Hour)
+
+	publisherStore := dssync.MutexWrap(datastore.NewMapDatastore())
+	rootLnk, err := test.Store(publisherStore, basicnode.NewString("hello world"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p := &head.Publisher{}
+	go p.Serve(ctx, publisher, "test")
```
gammazero, line 35: The context passed in here appears to control the lifetime of the server.  Yet the same context is passed to `QueryRootCid`, and that context should be used for cancellation of the function.  However, this test makes it look like the context is used for cancelation of either.

Ultimately, it seems that the context passed into `p.Serve` here does not really do anything since it does not cancel the inner goroutine.

p2p/protocol/head/head.go
```diff
@@ -0,0 +1,107 @@
...
+	l, err := gostream.Listen(host, deriveProtocolID(topic))
+	if err != nil {
+		return err
+	}
+
+	p.server = &http.Server{Handler: http.Handler(p)}
+	errChan := make(chan error)
+	go func() {
+		errChan <- p.server.Serve(l)
+	}()
```
gammazero, line 36: I think that instead of creating a go routine, you do:
```go
return p.server.Serve(l)
``` 
and let the caller decide if they want to wait in a goroutine, it will be simpler.  Since this function is always called
```go
go p.Server(...)
```
I do not see the point of creating an internal goroutine and waiting on it.  This also means that the context becomes meaningless since it is not passed into `p.server.Serve()` to cancel that call.

p2p/protocol/head/head.go
```diff
@@ -0,0 +1,107 @@
...
+	gostream "github.com/libp2p/go-libp2p-gostream"
+)
+
+var log = logging.Logger("go-legs-head")
+
+func deriveProtocolID(topic string) protocol.ID {
+	return protocol.ID("/legs/head/" + topic + "/0.0.1")
+}
+
+func (p *Publisher) Serve(ctx context.Context, host host.Host, topic string) error {
```
gammazero, line 26: Passing in a context here is confusing when there is a `Close()` function.  To shutdown the publisher, should the context be canceled, or should `Close` be called?  Can we just have one or the other?

I would prefer that we only use contexts for the cancellation of a function call that may block, not for controlling the lifetime of a service.

subscribe.go
```diff
@@ -268,6 +271,16 @@ func (ls *legSubscriber) Sync(ctx context.Context, p peer.ID, c cid.Cid, s ipld.
 	v := Voucher{&c}
 	var ulOnce sync.Once
 
+	if cid.Undef.Equals(c) {
```
gammazero, line 274: Prefer:
```go
if c == cid.Undef {
```

gammazero (commented):
Have a few comments, mostly related to the non-functional use of context passed into the `Serve` function.

p2p/protocol/head/head.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/libp2p/go-libp2p-core/protocol"
+	gostream "github.com/libp2p/go-libp2p-gostream"
+)
+
+const closeTimeout = 30 * time.Second
+
+var log = logging.Logger("go-legs-head")
+
+func deriveProtocolID(topic string) protocol.ID {
+	return protocol.ID("/legs/head/" + topic + "/0.0.1")
```
masih, line 25: If `0.0.1` represents the version would it make sense to move that after `/legs/`?

MarcoPolo: I don't think anything technically prohibits the version being in a certain place (or even not being there at all). There is a [helper](https://github.com/libp2p/go-libp2p-core/blob/master/helpers/match.go#L16) that expects the version to be in the last bit, but that's only if you use `SetStreamHandlerMatch`. In this case we don't, we use the simpler `SetStreamHandler`.

And even if we did use `SetStreamHandlerMatch` we could write our own matcher that got the version from wherever.

But I'm not convinced (either way) what is the correct thing here. My intuition is that the protocol id should serve to quickly refine if another protocol id is compatible and each part expands on the previous part (this is why the version isn't first). The sooner we can figure out if we are a compatible protocol the better.

In this case if the topic is different, we know this isn't a compatible protocol (regardless if the version matches). If the version were sooner, we would be version matching, then realize the topic is different at the end (possibly wasting some compute in version matching when we would have know the topic was incompatible anyways).

What's a little weird about this I think, and why it feels like it would make sense to move the version to before the topic, is that the topic string is part of the protocol id. This is a little weird in that a topic itself doesn't change the protocol at all. It's used as part of the protocol ID so that we have an unique stream per topic. It would almost make sense to think of this as the stream id instead.

p2p/protocol/head/head.go
```diff
@@ -0,0 +1,99 @@
...
+		return
+	}
+
+	p.rl.RLock()
+	defer p.rl.RUnlock()
+	out, err := json.Marshal(p.root.String())
+	if err != nil {
+		log.Infow("failed to serve root", "err", err)
+		http.Error(w, "", http.StatusInternalServerError)
+	} else {
```
gammazero, line 83: nit: return after `http.Error` and no else... since there is nothing to do after `http.Error`, so less chance of accidentally not returning.

p2p/protocol/head/head_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	cid, err := head.QueryRootCid(ctx, client, "test", publisher.ID())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !cid.Equals(rootLnk.(cidlink.Link).Cid) {
+		t.Fatalf("didn't get expected cid. expected %s, got %s", rootLnk, cid)
+	}
+}
```
gammazero, line 50: Should add negative test to make sure correct error response when not found.

gammazero (approved):

MarcoPolo:
@gammazero what do you think about merging this next week? I'm not sure what is involved in actually deploying this change so I'd love to pair on that next week.

willscott:
Is there a reason to wait that long? we should be able to merge & tag in a pretty light way way - unless there's more things that come up in a review, not reason to wait a week

MarcoPolo:
Sure. I was only thinking next week since it's the end of the bootcamp stuff. But if it's straightforward maybe we can do it in the colo today?

gammazero:
I think it is ready for merge now.  Before updating storetheindex to use the latest, test storetheindex locally first using the update.

PR #33, opened 2021-11-24
Use consistent logic when wrapping given selector sequence

masih:
In both constructor and explicit Sync behave consistently:
 - use the selector given as what it is: selection sequence that is
   wrapped in a custom selector.
 - fall back on the default selector if one is not provided.

Add docs to explain behaviour.

### DIFF
--- a/http/subscribe.go
+++ b/http/subscribe.go
@@ -29,9 +29,19 @@ var defaultPollTime = time.Hour
 
 var log = logging.Logger("go-legs-http")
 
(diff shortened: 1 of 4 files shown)


willscott:
I think you're missing the 'ExploreRecursiveWithStopNode' definition?

willscott:
cool. this is fine. can you also make it do the same thing in the `http/` package?

willscott (approved):

PR #36, opened 2021-11-28
Add message originator addresses to peerstore

gammazero:
Put the message originator's addresses into the pubsub message.  When this message is received by go-legs, the addresses are added to the peerstore.  This allows datatransfer to make a connection back to the provider to get advertisements and content.  This is necessary when the provider and indexer are not directly connected, but are connected through intermediate nodes.

### DIFF
--- a/message.go
+++ b/message.go
@@ -1,0 +1,63 @@
+package legs
+
+import (
+	"bytes"
+
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multiaddr"
+	"github.com/multiformats/go-varint"
+)
+
+func encodeMessage(c cid.Cid, addrs []multiaddr.Multiaddr) []byte {
+	// Get size
+	var size, maxVarintLen int
+	for i := range addrs {
+		addrBytes := addrs[i].Bytes()
+		varintLen := varint.UvarintSize(uint64(len(addrBytes)))
+		if varintLen > maxVarintLen {
+			maxVarintLen = varintLen
+		}
+		size += len(addrBytes) + varintLen
+	}
+
+	vibuf := make([]byte, maxVarintLen)
+	var b bytes.Buffer
+	b.Grow(c.ByteLen() + size)
+	// CID already contains encoded length, so no need to add length to data.
+	c.WriteBytes(&b)
+
+	for i := range addrs {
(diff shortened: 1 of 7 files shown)


message.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"bytes"
+
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multiaddr"
+	"github.com/multiformats/go-varint"
+)
+
+func encodeMessage(c cid.Cid, addrs []multiaddr.Multiaddr) []byte {
```
willscott, line 11: would it make sense to define a schema and cbor encode the data, rather than doing it by hand?

subscribe.go
```diff
@@ -168,51 +171,65 @@ func (ls *legSubscriber) watch(ctx context.Context, sub *pubsub.Subscription) {
...
+			continue
+		}
+		v := Voucher{&c}
+
+		// Add the message originator's address to the peerstore.  This
+		// allows a connection, back to that provider that sent the
+		// message, to retrieve advertisements.
+		peerStore := ls.host.Peerstore()
+		if peerStore != nil {
+			peerStore.AddAddrs(src, addrs, 5*time.Minute)
```
willscott, line 211: make a constant or var for time

i wonder if we want to keep known addresses for the peer around for the life of a subscriber so that if an explicit sync is called we know we have them?

willscott (approved):

subscribe.go
gammazero: If we do explicit sync, that means that we already know about the provider and that it is registered with the indexer and its address is known. So for manual sync, the provider address could be put into the peerstore at that time.

gammazero: Used the existing const `peerstore.ProviderAddrTTL` which seems all too appropriate.

message.go
gammazero: cbor_gen does not support slices, which I need to encode `addrs`, so I would need to write my own cbor encoding.  I was also worried about keeping the message as small as possible for getting spread via gossip. I did define a message datatype.  LMK if you would still prefer a cbor or other type of encoder.

willscott: not blocking, but we may eventually want to re-visit

PR #42, opened 2021-12-07
Fix race conditions

gammazero:
- Fix one race condition when starting the `head.Publisher` server.
- Fix multiple race conditions in tests where connections need to be created after gossip pubsub is created. This is so that connecting hosts can see that the destination host has pubsub, and will publish messages.

Note: The sleep to wait for the gossip mesh to establish is still needed, but the above fixes a race that when hit would always fail regardless of the gossip mesh wait.

### DIFF
--- a/message.go
+++ b/message.go
@@ -49,7 +49,7 @@ func decodeMessage(data []byte) (message, error) {
 	data = data[n:]
 
 	// Read multiaddrs if there is any more data in message data.  This allows
-	// backward-compatability with publishers that do not supply addres data.
+	// backward-compatability with publishers that do not supply address data.
 	var addrs []multiaddr.Multiaddr
 	for len(data) != 0 {
 		val, n, err := varint.FromUvarint(data)
--- a/message.go
+++ b/message.go
@@ -48,6 +48,8 @@ func decodeMessage(data []byte) (message, error) {
 	}
 	data = data[n:]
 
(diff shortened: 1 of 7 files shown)


message.go
```diff
@@ -48,6 +48,8 @@ func decodeMessage(data []byte) (message, error) {
 	}
 	data = data[n:]
 
+	// Read multiaddrs if there is any more data in message data.  This allows
+	// backward-compatability with publishers that do not supply addres data.
```
adlrocha, line 52:
```suggestion
	// backward-compatability with publishers that do not supply address data.
```

adlrocha, line 52:
```suggestion
	// backward-compatability with publishers that do not supply address data.
```

adlrocha (approved):

masih (approved):

p2p/protocol/head/head.go
```diff
@@ -28,7 +42,6 @@ func (p *Publisher) Serve(host host.Host, topic string) error {
 		return err
 	}
 
-	p.server = &http.Server{Handler: http.Handler(p)}
```
MarcoPolo, line 31: Yeah this is a bit subtle. In general we shouldn't be mutating the object that can be shared across goroutines without a mutex/other-sync-primitive. And we can do this part in the New func.

MarcoPolo: Thanks for making this change! 👍

PR #45, opened 2021-12-08
Create a new pubsub message broker for improved go-legs functionality

gammazero:
### New LegBroker type
The new LegBroker replaces both the `LegMultiSubscriber` and `LegSubscriber`.  It offers more scalability since there are no longer separate pubsub subscribers, and pairs of persistent goroutines, for each message source.  Instead, the new broker has only two persistent goroutines (for reading pubsub messages and for distributing events) and each handles messages in a goroutine that only lasts for the duration of handling the message.  The broker has improved logging and simplified event notification.  Decisions on whether to allow or reject messages for certain peers can now be decided by an external mechanism.

### Reorg repo
Other than the broker, there are two main packages now:
- `dtsync`: Code for syncing using go-data-transfer, and publishing over gossib pubsub
- `httpsync`: Code for syncing over http, and serving latest head via http

The old subscriber, and multisubscriver are still present, and can be removed if they are no longer needed as a reference implementation.

### Old code not removed yet
Even though the broker is intended to replace the `LegSubscriber` and `LegMultiSubscriber`, those bodies of code and their associated tests have been left in the repo for the purpose of comparison.  This is for reviewers and testers to ensure that the broker is capable of supporting all the use cases supported by the code it replaces.

**Do not merge this PR until the integration work is completed in `storetheindex`.**


broker.go
```diff
@@ -0,0 +1,584 @@
...
+
+	if !updateLatest {
+		return nil
+	}
+
+	// Update latest head seen.
+	// NOTE: This is not persisted anywhere. Is the top-level user's
+	// responsibility to persist if needed to initialize a
+	// partiallySynced subscriber.
+	h.latestSync = cidlink.Link{Cid: c}
```
masih, line 566: Can I ask you to document why we are not locking the `syncMutex` , and why are we not using `setLatestSync` instead please.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+func (lb *LegBroker) GetLatestSync(peerID peer.ID) ipld.Link {
+	lb.handlersMutex.Lock()
+	defer lb.handlersMutex.Unlock()
+
+	hnd, ok := lb.handlers[peerID]
+	if !ok {
+		return nil
+	}
+
+	return hnd.latestSync
```
masih, line 183: Could there be a race condition for reading this field via `GetLatestSync` at broker level and at handler level?

broker_legs_test.go
```diff
@@ -0,0 +1,240 @@
...
+import (
+	"context"
+	"io/ioutil"
+	"testing"
+	"time"
+
+	"github.com/ipfs/go-datastore"
+	dssync "github.com/ipfs/go-datastore/sync"
+	gsimpl "github.com/ipfs/go-graphsync/impl"
+	gsnet "github.com/ipfs/go-graphsync/network"
```
masih, line 12: Just so I learn the convention, why these imports are grouped together?

masih (approved):
Before removing the old implementation can I ask you to make sure the new version has logging in at least as many places as the old one please?

masih:
> Do not merge this PR until the integration work is completed in storetheindex.

Is it worth marking this as a draft PR?

MarcoPolo:
>  It offers more scalability since...

Not that I doubt it since the change seems like it should make perf better, but did you run any tests, and what did the tests say?

broker_legs_test.go
gammazero: This was just copy and past from `legs_test.go`.  The only convention I follow is grouing stdlib packages and other packages.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+// LegBroker is a message broker for pubsub messages.  LegBroker creates a
+// single pubsub subscriber that receives messages from a gossip pubsub topic,
+// and creates a stateful message handler for each message source peer.  An
+// optional external policy determines whether to allow or deny handling
+// messages from specific peers.
+//
+// Messages to separate peers are handled concurrently, and multiple messages
+// to the same peer are handled serially.  If a handler is busy handling a
+// message, and more messages arrive for the same peer, then the last message
+// replaces the previous unhandled message to avoid having to maintain queues
```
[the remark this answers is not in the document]
willscott: This seems to be the intention of the library as a whole. More verbose comments at this level is also at risk of not being read by users. This seems like an appropriate level of detail.

broker.go
```diff
@@ -0,0 +1,604 @@
...
+	// putSyncDone is a function supplied by the LegBroker for the handler
+	// to give LegBroker the channel to send sync done notification to.
+	putSyncDone func(cid.Cid, chan<- struct{})
+	// syncMutex serializes the handling of individual syncs
+	syncMutex sync.Mutex
+}
+
+// NewLegBroker creates a new LegBroker, and will process messages that are allowed by given peerPolicy.
+func NewLegBroker(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topic string, dss ipld.Node, addrTTL time.Duration, peerPolicy PeerPolicyHandler) (*LegBroker, error) {
+	ctx, cancelAll := context.WithCancel(context.Background())
```
willscott, line 117: maybe we should take a context in the constructor?

broker.go
```diff
@@ -0,0 +1,604 @@
...
+		return nil, err
+	}
+
+	psub, err := pubsubTopic.Subscribe()
+	if err != nil {
+		cancelAll()
+		return nil, err
+	}
+
+	dt, gs, tmpDir, err := makeDataTransfer(ctx, host, ds, lsys)
```
willscott, line 131: forcing an all-or-nothing constructor seems problematic here: we're using an existing data transfer and pubsub in the context of integration of index-provider. there needs to be a way to build a this broker (startup the background process) over such an existing setup.

broker.go
```diff
@@ -0,0 +1,604 @@
...
+// Messages to separate peers are handled concurrently, and multiple messages
+// to the same peer are handled serially.  If a handler is busy handling a
+// message, and more messages arrive for the same peer, then the last message
+// replaces the previous unhandled message to avoid having to maintain queues
+// of messages.  Handlers do not have persistent goroutines, but start a new
+// goroutine to handle a single message.
+type LegBroker struct {
+	dtManager   dt.Manager
+	gsExchange  graphsync.GraphExchange
+	tmpDir      string
```
willscott, line 53: why does the broker have a temp dir? this seems to be only used for tests and not a property of this code itself?

broker.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+// OnChange creates a channel that receives change notifications, and adds that
+// channel to the list of notification channels.
+//
+// Calling the returned cancel function removed the notification channel from
+// those the receive change notifications, and closes the channel to allow any
+// waiting goroutines to stop waiting on the channel.  For this reason, channel
+// readers should check for channel closure.
+func (lb *LegBroker) OnChange() (<-chan ChangeEvent, context.CancelFunc) {
```
willscott, line 244: If there's a single 'onchange' here, how do we incorporate non-advertising providers?

I wonder if there shouldn't be two layers of abstraction:
One object watching pubsub and auto-discovering providers to subscribe to, that can easily feed additions and removals into another watcher that emits changes across the collection of known providers. Then a user can manually add a provider in addition to using the auto discovery in a conceptually reasonable way.

willscott (commented):
you should remove the multi subscribe file if we go this route.

MarcoPolo (commented):
Just some questions on my initial pass. Hope to do a more thorough pass tomorrow.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+}
+
+// ChangeEvent notifies an OnChange reader that a specified peer completed a sync.
+type ChangeEvent struct {
+	Cid    cid.Cid
+	PeerID peer.ID
+}
+
+// Defined tells whether the ChangeEvent holds any data.
+func (ce ChangeEvent) Defined() bool {
```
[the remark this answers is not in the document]
gammazero: Agree, and even that should not occur, except when the event channel closes.In that case the reader should check for channel closed instead of whether the event is defined.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+	// LegBroker for distribution to OnChange readers.
+	distEvents chan<- ChangeEvent
+	latestSync ipld.Link
+	msgChan    chan cid.Cid
+	// peerID is the ID of the peer this handler is responsible for.
+	peerID peer.ID
+	// putSyncDone is a function supplied by the LegBroker for the handler
+	// to give LegBroker the channel to send sync done notification to.
+	putSyncDone func(cid.Cid, chan<- struct{})
+	// syncMutex serializes the handling of individual syncs
```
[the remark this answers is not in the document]
gammazero: That handles the serialization of the sync for an individual handler.  That includes changes to latestSync, but also the serialization of the sync operation itself.

broker.go
gammazero: I really wanted to get rid of that because I do not want to use a Context to control the lifetime of a service. It has a `Close()` function for that. There is no need to cancel the constructor, and I do not anyone to think they should use a Context to control the service lifetime.  This is actually a pattern I want to rid our software of, and only use constructors for function cancellation.  If we to shutdown a service, then provide a `chan struct{}` to close or provide a `Close()` function, or if possible make a service that runs for the life of a process crash-only shutdown, where crash is a perfectly safe way to stop the service.

broker.go
gammazero: The events here do not come from watching pubsub, but rather watching events generated by the datatransfer manager.  The channel will also receive events from providers that are manually synced to the latest, as well as those auto-discovered.  A new provider can be added manually by calling `Broker.Sync()` and passing the peer ID of the provider to discover.

willscott: If i call `Sync` with a multiaddr that's an http multiaddr, i don't see where that would get delegated to the HTTP subscriber package?

legs_test.go
```diff
@@ -197,11 +200,21 @@ func TestSetAndFilterPeerPolicy(t *testing.T) {
 	np := basicnode.Prototype__Any{}
 	nb := np.NewBuilder()
 	ma, _ := nb.BeginMap(2)
-	ma.AssembleKey().AssignString("hey")
-	ma.AssembleValue().AssignString("it works!")
-	ma.AssembleKey().AssignString("yes")
-	ma.AssembleValue().AssignBool(true)
-	ma.Finish()
+	if err = ma.AssembleKey().AssignString("hey"); err != nil {
```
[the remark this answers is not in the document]
gammazero: It does not provide any additional functionality, only syntax that may be preferable to some.  I would prefer that be done in a separate PR if you feel it would be a useful improvement.

broker_sync_test.go
```diff
@@ -0,0 +1,361 @@
...
+	lb, err := NewLegBroker(dstHost, dstStore, dstLnkS, testTopic, nil, 0, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := srcHost.Connect(context.Background(), dstHost.Peerstore().PeerInfo(dstHost.ID())); err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
gammazero: Yes, that has been something that has been requested for a long time so that we can avoid guessing at sleep times (really, sleeps should never be in tests if it can be avoided).  I think Daniel may have put a fix in a later version of libp2p-pubsub, but we need to see if we can upgrade to that version yet.  Another PR.

broker.go
```diff
@@ -0,0 +1,604 @@
...
+			// Query the peer for the latest CID
+			var err error
+			c, err = head.QueryRootCid(ctx, lb.host, lb.topic.String(), peerID)
+			if err != nil {
+				log.Errorw("Cannot get root for sync", "err", err, "peer", peerID)
+				return
+			}
+			log.Debugw("Sync queried root CID", "peer", peerID, "cid", c)
+			if selSeq == nil {
+				// Update the latestSync only if no CID and no selector given.
```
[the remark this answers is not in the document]
gammazero: Copied from previous logic.  Here is I think the reasoning is:
- Do not update `latestSync` if given a specific CID, because it may not actually be the latest, and we may already have something later.
- Do not update `latestSync` if given a custom selector because who knows what it selects.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+// optional external policy determines whether to allow or deny handling
+// messages from specific peers.
+//
+// Messages to separate peers are handled concurrently, and multiple messages
+// to the same peer are handled serially.  If a handler is busy handling a
+// message, and more messages arrive for the same peer, then the last message
+// replaces the previous unhandled message to avoid having to maintain queues
+// of messages.  Handlers do not have persistent goroutines, but start a new
+// goroutine to handle a single message.
+type LegBroker struct {
```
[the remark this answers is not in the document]
gammazero: I was trying to follow the naming already present but would like to change this.  I think I can drop the `Leg` prefix as that stutters given that the package is named `legs`.

I named it Broker because it receives messages from various peers over pubsub and then relays the messages to separate the handler for each peer.

Other names considered:
Dispatcher, Receiver, Router, Communicator

Something called `legs.Broker` seems like the most fun.

broker.go
```diff
@@ -0,0 +1,604 @@
...
+				updateLatest = true
+			}
+		}
+
+		if ctx.Err() != nil {
+			log.Errorw("Sync canceled", "err", ctx.Err(), "peer", peerID)
+			return
+		}
+
+		hnd.syncMutex.Lock()
```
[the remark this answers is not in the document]
gammazero: Yes, thanks.
Fixed.

broker.go
```diff
@@ -0,0 +1,604 @@
...
+// in order to avid re-syncing of content that may have previously been synced.
+//
+// The selector sequence, selSec, can optionally be specified to customize the
+// selection sequence during traversal.  If unspecified, the subscriber's
+// default selector sequence is used.
+//
+// Note that the selector sequence is wrapped with a selector logic that will stop traversal when
+// the latest synced link is reached. Therefore, it must only specify the selection sequence itself.
+//
+// See: ExploreRecursiveWithStopNode.
```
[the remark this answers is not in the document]
gammazero: Yes to the first part.  We want to walk back as far as is reasonable, possibly to the beginning of a provider's records, to get all content.  A new indexer may come online at any time and will want to get content that has already been published.  Also handles missed messages.

The second part is because that is how we make IPLD/graphsync walk a DAG (a chain in this case).

broker.go
```diff
@@ -0,0 +1,604 @@
...
+	default:
+	}
+
+	// Put new message on channel.  This is necessary so that if multiple
+	// messages arrive while this handler is already syncing, the messages are
+	// handled in order, regardless of goroutine scheduling.
+	h.msgChan <- c
+
+	go func() {
+		// Wait for any previous messages to be handled.
```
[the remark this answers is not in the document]
gammazero: Reworded to say:
```go
// Wait for this handler to become available.
```

broker.go
```diff
@@ -0,0 +1,604 @@
...
+// the peer, and results in the messages being rejected.
+type PeerPolicyHandler func(peerID peer.ID) (bool, error)
+
+// LegBroker is a message broker for pubsub messages.  LegBroker creates a
+// single pubsub subscriber that receives messages from a gossip pubsub topic,
+// and creates a stateful message handler for each message source peer.  An
+// optional external policy determines whether to allow or deny handling
+// messages from specific peers.
+//
+// Messages to separate peers are handled concurrently, and multiple messages
```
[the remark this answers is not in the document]
gammazero: Yes, good catch.  Fixed.

broker.go
```diff
@@ -0,0 +1,604 @@
...
+
+		peerPolicy: peerPolicy,
+		handlers:   make(map[peer.ID]*handler),
+		inEvents:   make(chan ChangeEvent, 1),
+	}
+
+	lb.unsubEvents = dt.SubscribeToEvents(lb.onEvent)
+
+	// Start watcher to read pubsub messages.
+	go lb.watch(ctx)
```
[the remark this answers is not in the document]
gammazero: I completely agree.  However, inside `watch`, I need to call `psub.Next(ctx)` and pass it a context to cancel when  LegBroker is closed.  The cancel function needs to be called from `Close()`, so this context is created in the constructor, and certainly not in the goroutine.

For stopping the `distribute` goroutine I will close its read channel.

Done

broker_legs_test.go
```diff
@@ -0,0 +1,240 @@
...
+		t.Fatal(err)
+	}
+
+	select {
+	case <-time.After(time.Second * 3):
+	case _, open := <-watcher:
+		if open {
+			t.Fatal("something was exchanged, and that is wrong")
+		}
+	}
```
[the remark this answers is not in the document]
gammazero: Done.

broker.go
gammazero: That is the temp dir from the datatransfer manager, and it needs to be removed in `Close()`

willscott: what about cases where the broker doesn't own the data transfer? or are we assuming on the subscribe side that it always will?

this means i couldn't use this in something like the retrieval client where there's an active data transfer shared on the host that's also being used for active data fetches?

broker.go
gammazero: I have changed this to let existing `datatransfer.Manager` and `pubsub.Topic` to be given as options.

broker.go
gammazero: Fixed

broker.go
```diff
@@ -0,0 +1,584 @@
...
+	if err != nil {
+		cancelAll()
+		return nil, err
+	}
+
+	if addrTTL == 0 {
+		addrTTL = defaultAddrTTL
+	}
+
+	ctx, cancelBrkr := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
gammazero: It was originally done because I thought I needed to shut some things down separately from the pubsub and datatransfer, but that was because of a problem that was fixed elsewhere.  This does not need to be separated now.

Fixed.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+	peerID peer.ID
+	// putSyncDone is a function supplied by the LegBroker for the handler
+	// to give LegBroker the channel to send sync done notification to.
+	putSyncDone func(cid.Cid, chan<- struct{})
+	// syncMutex serializes the handling of individual syncs
+	syncMutex sync.Mutex
+}
+
+// NewLegBroker creates a new LegBroker, and will process messages that are allowed by given peerPolicy.
+func NewLegBroker(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topic string, dss ipld.Node, addrTTL time.Duration, peerPolicy PeerPolicyHandler) (*LegBroker, error) {
```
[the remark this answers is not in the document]
gammazero: @masih Sure, I will add options.  I was just trying to keep it closer to LegMultisubscriber, but yeah... options would be good.  Done.

@MarcoPolo With some of the changes, it cannot be a drop-in.  It is close enough to make it tempting to try and do that, but I think that would detract from clarity.  I do not see the need for defining an interface, aside from test mocking, since we are only supplying on implementation.  If there was a consumer of multiple implementations, that consumer can define that interface.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+	// cancelBroker cancels broker goroutines
+	cancelBrkr context.CancelFunc
+	// cancelAll cancels pubsub and datatransfer and broker goroutines
+	cancelAll context.CancelFunc
+}
+
+// ChangeEvent notifies an OnChange reader that a specified peer completed a sync.
+type ChangeEvent struct {
+	Cid    cid.Cid
+	PeerID peer.ID
```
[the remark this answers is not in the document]
gammazero: Done

broker.go
```diff
@@ -0,0 +1,584 @@
...
+	outEventsMutex sync.Mutex
+
+	// cancelBroker cancels broker goroutines
+	cancelBrkr context.CancelFunc
+	// cancelAll cancels pubsub and datatransfer and broker goroutines
+	cancelAll context.CancelFunc
+}
+
+// ChangeEvent notifies an OnChange reader that a specified peer completed a sync.
+type ChangeEvent struct {
```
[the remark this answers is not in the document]
gammazero: Makes sense.  I was originally thinking there may be different event types, but that has not materialized.

Done.

broker.go
```diff
@@ -0,0 +1,584 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+// defaultAddrTTL is the default amount of time that addresses discovered from
+// pubsub messages will remain in the peerstore.  This is twice the default
+// provider poll interval.
+const defaultAddrTTL = 48 * time.Hour
+
+var ErrSourceNotAllowed = errors.New("message source not allowed by policy")
```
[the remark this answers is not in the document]
gammazero: Done.

broker.go
gammazero: If they supply their own `dtManager`, then tmpDir remains `""` and no attempt is made to delete it.

broker.go
gammazero: There is no code to do that currently.

willscott:
Also make sure to update the readme to reflect the new flow

utils.go
```diff
@@ -120,12 +124,25 @@ func makeDataTransfer(ctx context.Context, host host.Host, ds datastore.Batching
...
+		if err = dt.Start(ctx); err != nil {
+			log.Errorf("Failed to start datatransfer: %s", err)
+			return nil, nil, "", err
+		}
+	}
+
+	// Wait for datatrnasfer to be ready.
+	dtReady := make(chan error)
+	dt.OnReady(func(e error) {
+		dtReady <- e
```
MarcoPolo, line 142: Does this work if you pass in a dt.Manager that is already started? I think not.

broker.go
```diff
@@ -0,0 +1,645 @@
...
+}
+
+// NewBroker creates a new Broker that process pubsub messages.
+func NewBroker(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topic string, dss ipld.Node, options ...Option) (*Broker, error) {
+	cfg := config{
+		addrTTL: defaultAddrTTL,
+	}
+	err := cfg.apply(options)
+	if err != nil {
+		panic(err.Error())
```
MarcoPolo, line 129: Why panic?

broker.go
```diff
@@ -0,0 +1,645 @@
...
+	}
+	lb.handlersMutex.Unlock()
+
+	return hnd.getLatestSync()
+}
+
+// SetLatestSync sets the latest synced CID for a specified peer.  If there is
+// no handler for the peer, then one is created without consulting any
+// AllowPeerFunc.
+func (lb *Broker) SetLatestSync(peerID peer.ID, latestSync cid.Cid) error {
```
MarcoPolo, line 201: What do you think abouut this func returning an error if you pass in `latestSync == cid.Undef`? This way the caller gets an error rather this it silently doing nothing.

MarcoPolo (commented):
PR makes sense 👍 . Anything you'd like me to do with this PR?

utils.go
gammazero: Thank you. Fixed

broker.go
gammazero: Was thinking to ensure this was caught as a programming error.  Probably better just to return the error.

broker.go
gammazero: I was thinking that passing `cid.Undef` provided a way to create a handler without actually doing a sync.  That might be useful for bypassing the AllowPeer without doing a sync.

However, I cannot really think of a use case where it would not be preferable to update the AllowPeer instead or just do an explicit sync, so passing `cid.Undef` is more likely the result of an error and should therefore be treated as such.

broker.go
```diff
@@ -0,0 +1,645 @@
...
+	})
+	return err
+}
+
+func (lb *Broker) doClose() error {
+	lb.unsubEvents()
+	lb.psub.Cancel()
+
+	var errs error
+	err := lb.dtManager.Stop(context.Background())
```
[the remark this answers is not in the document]
gammazero: Or, no... That would be bad. Wow, cannot believe I missed that.

Same situation for `lb.topic`... don't close it if it was passed in.

Fixed

publish.go
```diff
@@ -71,16 +71,15 @@ func NewPublisherFromExisting(ctx context.Context,
...
 		cid:   c,
 		addrs: lp.host.Addrs(),
 	}
-	err1 := lp.topic.Publish(ctx, encodeMessage(msg))
-	err2 := lp.headPublisher.UpdateRoot(ctx, c)
-	if err1 != nil {
-		return err1
-	}
-	return err2
+	return lp.topic.Publish(ctx, encodeMessage(msg))
```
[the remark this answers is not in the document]
gammazero: The idea is that if `lp.headPublisher.UpdateRoot(ctx, c)` returns an error, then headPublisher would be out of sync with what is published on pubsub, and that might cause a problem.  Maybe its, not a realistic problem, and it should still publish on pubsub anyway.
Fixed.

gammazero:
@MarcoPolo The next piece I need to add to this is to combine the `http.Subscriber` with the Broker so that the Broker can handle http syncs and the same event notification channel to send `SyncFinished` events when an http sync finishes.  Also, I do not want every http peer to have a separate persistent goroutine, but instead, the broker should dispatch the http sync request to the peer-specific handler to handle the sync.

I should have that done tomorrow, and that should conclude the legs development.  Then just need to replace the legs subscriber stuff in storetheindex with the broker.

masih (approved):
There is a lot of code here for me to review. 

Subjective cosmetics comment:
* Based on a quick glance I see a few additional concepts like `Sync`, `Syncer`, `Broker` etc. which makes me wonder if it is possible to reduce them to fewer concepts. The more the concepts the higher the learning curve.

* I am also not sure if the package structure is most ergonomic when it comes to UX. 

Regardless, this PR is adding a lot of value and I think it's best to get it through to master sooner rather than later, and alter as needed iteratively

Thank you so much for this @gammazero 🍻

.gitignore
```diff
@@ -0,0 +1,4 @@
+storetheindex
+*~
```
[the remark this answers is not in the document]
gammazero: Done, thanks.

broker/broker.go
```diff
@@ -0,0 +1,607 @@
+package broker
```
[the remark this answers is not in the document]
gammazero: Move Broker into legs package.

broker.go
gammazero: This is done here: https://github.com/filecoin-project/go-legs/pull/45/files#diff-879d7ff817ef6271a190861e67ab0ff3a2751136cedff764b3f680565540dc2bR329-R330

broker.go
gammazero: This is now handled: https://github.com/filecoin-project/go-legs/pull/45/files#diff-6fc0e45737bd8bdc5c8320e42995cf43fab5825133af1adeba404244cb32fed4R69-R71

broker_sync_test.go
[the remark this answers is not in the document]
gammazero: Replaced waiting for mesh using time.Sleep() with a function that can be updated when a better way is available.

broker.go
```diff
@@ -0,0 +1,606 @@
...
+// particular sync to complete.  Any OnSyncFinished readers will also get a
+// SyncFinished when the sync succeeds, but only if syncing to the latest using
+// the default selector.
+//
+// It is the responsibility of the caller to make sure the given CID appears
+// after the latest sync in order to avid re-syncing of content that may have
+// previously been synced.
+//
+// The selector sequence, selSec, can optionally be specified to customize the
+// selection sequence during traversal.  If unspecified, the subscriber's
```
MarcoPolo, line 310:
```suggestion
// selection sequence during traversal.  If unspecified, the broker's
```

broker.go
```diff
@@ -0,0 +1,606 @@
...
+// Sync performs a one-off explicit sync with the given peer for a specific CID
+// and updates the latest synced link to it.
+//
+// The Context passed in controls the lifetime of the background sync process,
+// so make sure the context has an appropriate deadline, if any.
+//
+// A SyncFinished channel is returned to allow the caller to wait for this
+// particular sync to complete.  Any OnSyncFinished readers will also get a
+// SyncFinished when the sync succeeds, but only if syncing to the latest using
+// the default selector.
```
MarcoPolo, line 303:
```suggestion
// the default selector and a `cid.Undef`.
```

But maybe this is specified by saying "syncing to the latest." so feel free to disregard if you think it's clear enough as is.

broker.go
```diff
@@ -0,0 +1,606 @@
...
+	out := make(chan SyncFinished, 1)
+	go func() {
+		defer close(out)
+
+		var updateLatest bool
+		if c == cid.Undef {
+			// Query the peer for the latest CID
+			c, err = syncer.GetHead(ctx)
+			if err != nil {
+				log.Errorw("Cannot get head for sync", "err", err, "peer", peerID)
```
MarcoPolo, line 362: nit (feel free to skip), can we use `log.With` above and automatically get the peer id everywhere we log in this func?

httpsync/sync.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (s *Syncer) walkFetch(ctx context.Context, rootCid cid.Cid, sel selector.Selector) error {
+	getMissingLs := cidlink.DefaultLinkSystem()
+	// trusted because it'll be hashed/verified on the way into the link system when fetched.
+	getMissingLs.TrustedStorage = true
+	getMissingLs.StorageReadOpener = func(lc ipld.LinkContext, l ipld.Link) (io.Reader, error) {
+		r, err := s.sync.lsys.StorageReadOpener(lc, l)
+		if err != nil {
+			log.Errorw("Failed to get block read opener", "err", err, "link", l)
+			return r, nil
```
MarcoPolo, line 116: I'm confused by this. The old code had a check err == nil and returned `r, nil`. Now we're checking err != nil and return `r, nil`.

Seems to me that we want to return `r` if we were able to get the block from the storageReadOpener, no?

dtsync/publish.go
```diff
@@ -0,0 +1,161 @@
...
+				log.Errorf("Failed to stop datatransfer manager: %s", err)
+				errs = multierror.Append(errs, err)
+			}
+			if err = os.RemoveAll(lp.tmpDir); err != nil {
+				log.Errorf("Failed to remove temp dir: %s", err)
+				errs = multierror.Append(errs, err)
+			}
+		}
+
+		t := time.AfterFunc(shutdownTime, lp.cancelPubSub)
```
MarcoPolo, line 151: Why do we delay here?

MarcoPolo (commented):
My only blocker is understanding this: https://github.com/filecoin-project/go-legs/pull/45/files#r770114406

But besides that it looks good, once that's addressed I'll approve this PR.

Side note, sorry it took so long. This was essentially a whole repo refactor. But I think it's much better now. I especially like how the `Syncer`s now don't have to handle a `cid.Undef` input, it's handled at a higher level. Much cleaner 👍

MarcoPolo (approved):

dtsync/publish.go
gammazero: To give it time for `Close` to finish (shutdown gracefully) before canceling it.

MarcoPolo (approved):

httpsync/sync.go
gammazero: Yes.  Now returns r if `err == nil`

broker.go
gammazero: Fixed

broker.go
```diff
@@ -0,0 +1,606 @@
...
+
+	// Stop the distribution goroutine.
+	close(lb.inEvents)
+	return errs
+}
+
+// OnSyncFinished creates a channel that receives change notifications, and
+// adds that channel to the list of notification channels.
+//
+// Calling the returned cancel function removed the notification channel from
```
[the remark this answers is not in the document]
gammazero: Applied

MarcoPolo (approved):

broker.go
```diff
@@ -0,0 +1,609 @@
...
+	}
+
+	lb := &Broker{
+		dss:  dss,
+		host: host,
+		lsys: lsys,
+
+		addrTTL:   cfg.addrTTL,
+		psub:      psub,
+		topic:     pubsubTopic,
```
MarcoPolo, line 159: Should this be `cfg.topic` instead? Otherwise this will be nil if `cfg.topic` is set (via option)

MarcoPolo (changes requested):

broker.go
gammazero: yes

test/util.go
```diff
@@ -32,7 +32,7 @@ func WaitForMeshWithMessage(t *testing.T, topic string, hosts ...host.Host) []*p
 	meshFormed := false
 	defer func() {
 		if meshFormed {
-			fmt.Println("Mesh formed in", time.Since(now))
+			t.Log("Mesh formed in", time.Since(now))
```
MarcoPolo, line 35: 👍

MarcoPolo (approved):

gammazero:
Replaced by #57 to resolve complicated merge issues.

PR #47, opened 2021-12-09
Implement publisher admin HTTP server

masih:
...that allows querying the peer state, connect to given address and
re-publish the current head or a custom cid onto pubsub.

This PR is here for visibility and will not be merged into the master. Used for debugging purposes.

### DIFF
--- a/publish.go
+++ b/publish.go
@@ -2,7 +2,10 @@ package legs
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"strings"
+	"time"
 
 	dt "github.com/filecoin-project/go-data-transfer"
 	"github.com/filecoin-project/go-legs/p2p/protocol/head"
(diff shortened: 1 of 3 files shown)


publish.go
```diff
@@ -66,29 +79,244 @@ func NewPublisherFromExisting(ctx context.Context,
...
 	if err1 != nil {
 		return err1
+	} else if err2 != nil {
+		return err2
 	}
-	return err2
+	return err3
+}
+
+func (lp *legPublisher) startAdminServer(addr string) *http.Server {
```
willscott, line 139: This is not something that should be turned on by default as part of a legs publisher.

Maybe a http.mux for visibility could be provided in a utility package that an app can choose to link up, but the structure and choice of an http server really belongs in the application that uses the library for synchronization, not this library.

comparable wrappers like grpc / etc don't silently turn on an http admin servers when they're used.

masih: I do not intend to merge any of this; PR is in draft and created _purely_ for visibility.

masih: Added to PR description for clarity.

masih: Another reason this change is here and not in provider is because we can't upgrade lotus to the latest of index-provider until [this](https://github.com/filecoin-project/storetheindex/issues/130) is done.

masih:
This is now well out of date and needs rework considering the refactors in go-legs since this PR was created. Closing.

PR #55, opened 2021-12-16
Fix flaky tests with direct peers pubsub

MarcoPolo:
draft currently to give you the gist. If it looks good, I'll update all the tests to be less flaky:

## Summary
We can create the gossipsub pubsub object with the [pubsub.WithDirectPeers](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#WithDirectPeers) option.
>  all (valid) message unconditionally forwarded to them

This seems ideal for what we want in our test since we aren't trying to test the gossipsub implementation. So to enable that I added a new helper in test/util:

```go
// WaitForMeshWithMessage sets up a gossipsub network and sends a test message.
// Blocks until all other hosts see the first host's message.
func WaitForMeshWithMessage(t *testing.T, topic string, hosts ...host.Host) []*pubsub.Topic {
// ...
}
```

You can give it a topic name and a bunch of hosts, it will make sure the hosts connect to each other, create the pubsub object with the directpeer option, and create the topic. It will then send a test message on the topic and make sure the every other host gets the test message before returning. It uses the given topic to run this test so we know that that topic is ready to be used later on in the tests.

Broker already allows passing in a topic object via the Option. This adds something similar to dtsync publisher.

This mesh forms in ~50ms, so it can really speed up our tests (relative to sleeps in the seconds).

Tested against flakiness by using the `-count` option when running tests.

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -33,13 +33,6 @@ jobs:
         uses: protocol/multiple-go-modules@v1.2
         with:
           run: go test -v -coverprofile module-coverage.txt ./...
-      - name: Run tests (32 bit)
-        if: ${{ matrix.os != 'macos' }} # can't run 32 bit tests on OSX.
-        uses: protocol/multiple-go-modules@v1.2
-        env:
-          GOARCH: 386
-        with:
-          run: go test -v ./...
       - name: Run tests with race detector
         if: ${{ matrix.os == 'ubuntu' }} # speed things up. Windows and OSX VMs are slow
         uses: protocol/multiple-go-modules@v1.2
--- a/broker.go
+++ b/broker.go
@@ -156,8 +156,8 @@ func NewBroker(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topi
 
 		addrTTL:   cfg.addrTTL,
 		psub:      psub,
-		topic:     pubsubTopic,
-		topicName: pubsubTopic.String(),
+		topic:     cfg.topic,
+		topicName: cfg.topic.String(),
 		closing:   make(chan struct{}),
 		cancelps:  cancelPubsub,
 
--- a/dtsync/publish.go
+++ b/dtsync/publish.go
@@ -33,14 +33,27 @@ type publisher struct {
 const shutdownTime = 5 * time.Second
 
 // NewPublisher creates a new legs publisher
-func NewPublisher(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topic string) (*publisher, error) {
-	ctx, cancel := context.WithCancel(context.Background())
-
-	t, err := gpubsub.MakePubsub(ctx, host, topic)
+func NewPublisher(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topic string, options ...Option) (*publisher, error) {
+	cfg := config{}
+	err := cfg.apply(options)
 	if err != nil {
-		cancel()
 		return nil, err
 	}
+
+	var cancel context.CancelFunc
+	t := cfg.topic
+	if t == nil {
+		var ctx context.Context
+		ctx, cancel = context.WithCancel(context.Background())
+		t, err = gpubsub.MakePubsub(ctx, host, topic)
+		if err != nil {
+			cancel()
+			return nil, err
+		}
+	} else {
+		cancel = func() {}
+	}
+
 	dtManager, _, tmpDir, err := makeDataTransfer(context.Background(), host, ds, lsys, nil)
 	if err != nil {
 		cancel()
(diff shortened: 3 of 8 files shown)


dtsync/publish.go
```diff
@@ -33,14 +33,24 @@ type publisher struct {
...
+	ctx, cancel := context.WithCancel(context.Background())
+
+	t := cfg.topic
+	if t == nil {
+		t, err = gpubsub.MakePubsub(ctx, host, topic)
+		if err != nil {
+			cancel()
+			return nil, err
+		}
+	}
```
gammazero, line 52:
```suggestion
	var cancel context.CancelFunc
	t := cfg.topic
	if t == nil {
		var ctx context.Context
		ctx, cancel = context.WithCancel(context.Background())
		t, err = gpubsub.MakePubsub(ctx, host, topic)
		if err != nil {
			cancel()
			return nil, err
		}
	} else {
		cancel = func () {}
	}
```

dtsync/publish.go
```diff
@@ -60,6 +70,25 @@ func NewPublisher(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, t
 	}, nil
 }
 
+func NewPublisherWithTopic(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, t *pubsub.Topic) (*publisher, error) {
```
gammazero, line 73: Why is this needed if `NewPublisher` takes a `pubsub.Topic` as an option?

test/util.go
```diff
@@ -26,6 +33,95 @@ func WaitForMesh() {
 	time.Sleep(2 * time.Second)
 }
 
+// WaitForMeshWithMessage sets up a gossipsub network and sends a test message.
+// Blocks until all other hosts see the first host's message.
+func WaitForMeshWithMessage(t *testing.T, topic string, hosts ...host.Host) []*pubsub.Topic {
```
gammazero, line 38: Is it necessary for this function to create the slice of `pubsub.Topic`?  Wouldn't it be sufficient to just wait until message massing on the `startupCheck` topic was working?

gammazero (commented):
Look like a great fix.  You need to rebase this on my latest changes.
Have a couple of questions, see review.

test/util.go
```diff
@@ -15,9 +18,13 @@ import (
 	basicnode "github.com/ipld/go-ipld-prime/node/basic"
 	"github.com/libp2p/go-libp2p"
 	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/multiformats/go-multicodec"
 )
 
+const waitForMeshTimeout = 10 * time.Second
```
gammazero, line 26:
```suggestion
const waitForMeshTimeout = 2 * time.Second
```

test/util.go
MarcoPolo: yes.

What you suggest is what I tried initially, but it didn't work. It's because the `pubsub.Topic` comes from the `pubsub.PubSub` object, so in order for the topic to get the properties we want, we need to create it from this `pubsub.PubSub`.

dtsync/publish.go
MarcoPolo: my hack, I forgot to remove

MarcoPolo:
Cool I'll rebase, make sure this passes the tests, and clean up this code (It was mostly a PoC to get us on the same page). Then I'll request a review.

test/util.go
```diff
@@ -115,6 +117,11 @@ func WaitForMeshWithMessage(t *testing.T, topic string, hosts ...host.Host) []*p
 		select {
 		case <-done:
 		case <-time.After(waitForMeshTimeout):
+			if hackyRecursionLimit > 0 {
+				hackyRecursionLimit -= 1
+				return WaitForMeshWithMessage(t, topic, hosts...)
+			}
```
gammazero: Is there any reason to expect this to work over a longer wait time? Has that ever been observed?

MarcoPolo: no, I think if it doesn't complete in <1s then it will fail. Not sure why yet. Only happens in CI from what I can tell :(

.github/workflows/go-test.yml
```diff
@@ -33,13 +33,6 @@ jobs:
         uses: protocol/multiple-go-modules@v1.2
         with:
           run: go test -v -coverprofile module-coverage.txt ./...
-      - name: Run tests (32 bit)
```
MarcoPolo, line 36: I removed this since I don't think we actually need it. It adds a significant latency to run all the tests again in 32bit.

gammazero: The problems with is that the file is auto-generated, so next time there is a change to the CI stuff this gets undone.  Hence the comment at the top:
```go
# File managed by web3-bot. DO NOT EDIT.
```

I might be better to have explicit skips in the tests that fail, like what is done [here](https://github.com/filecoin-project/go-indexer-core/blob/main/store/pogreb/pogreb_test.go#L32) and [here](https://github.com/filecoin-project/go-indexer-core/blob/main/store/pogreb/pogreb_test.go#L87-L91)

broker.go
```diff
@@ -156,8 +156,8 @@ func NewBroker(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, topi
 
 		addrTTL:   cfg.addrTTL,
 		psub:      psub,
-		topic:     pubsubTopic,
-		topicName: pubsubTopic.String(),
+		topic:     cfg.topic,
+		topicName: cfg.topic.String(),
```
gammazero, line 160: 🎉

gammazero (approved):

PR #57, opened 2021-12-17
New Subscriber

gammazero:
The new `Subscriber` replaces both the `LegMultiSubscriber` and `LegSubscriber`. It offers more scalability since there are no longer separate pubsub subscribers and pairs of persistent goroutines for each message source. Instead, the new `Subscriber` has only two persistent goroutines (for reading pubsub messages and for distributing events).  Each message is handled in a goroutine, for each publisher, that only lasts for the duration of handling the message. The new `Subscriber` has improved logging and simplified event notification. Decisions on whether to allow or reject messages from publishers are now decided by an external mechanism.

Reorg repo
Other than the `Subscriber`, there are two main packages now:

`dtsync`: Code for syncing using go-data-transfer, and publishing over gossib pubsub
`httpsync`: Code for syncing over http, and serving latest head via http

An incoming block hook function can now be registered with the `Subscriber`, instead of having to register with `graphsync.GraphExchange` that the `Subscriber` is using.  This same incoming block hook is also called when a block arrives over HTTP during an explicit http sync.

This PR replace #45


interface.go
```diff
@@ -20,44 +18,6 @@ type Publisher interface {
 	Close() error
 }
 
-// LegSubscriber is an interface for watching a published dag.
-type LegSubscriber interface {
```
MarcoPolo, line 30: 👍

MarcoPolo (approved):

PR #58, opened 2021-12-18
[http] Sign head messages

MarcoPolo:
## Context

We want the head cid to be signed so that we know everything we pull from that cid + selector will be signed as well. In other words, having a signed head means blocks fetched will be transitively signed (yay content addressing schemes).

## Proposed solution

1. Create a `signedHead` struct that represents a signed head cid. It can be opened by the caller after verifying the identity. It has an associated IPLD schema. It encodes/decodes via dag-json.
2. Users/Callers use the encoded form of signedHead. So they create the opaque blob that is the encoded signedHead, and pass that to whoever wants the latest signed head. Readers will take that opaque blob of the encoded signedHead and "open it". They can only open it if the signature is valid.
3. The `signedHead` includes the public key used for signing. This is a convenience decision so that subscribers don't have to manage public key themselves, and instead need only deal with peer ids. Note, you can't extract the public key from a peer id since it is a hash of the peer's public key – not the actual key (except if it's encoded as an identity multihash, but this shouldn't be relied upon).
4. httpsync syncer will use the peerid it was given at initialization to validate if a head is properly signed.


httpsync/message.go
```diff
@@ -0,0 +1,96 @@
...
+// public key of the signer so the receiver can verify it and convert it to a
+// peer id. Note, the receiver is not required to use the provided public key.
+// Also note the signature is over the whole marshalled headmsg instead of just
+// a cidStr. This protects us from forgetting to sign something in the future if
+// we add to HeadMsg.
+type SignedHeadMsgEnvelope struct {
+	HeadMsg []byte
+	Sig     []byte
+	PubKey  []byte
+}
```
willscott, line 26: consider this data struct set up via ipld bind node. & serialized with ipld / dag-json.

i think the wire format change is the cid instead of just being a string would end up dag-json serialized to `{"/":{"cid"}}`, it means that these signed objects can get dumped into the datastore and synced subsequently without special care.

gammazero:
Unfortunately, so much has changed in the repo since when the PR was created that this needs a rebase or reapplying the changes to the current repo to make it really reviewable.

MarcoPolo:
@gammazero updated

httpsync/message.go
MarcoPolo: Can you elaborate a bit more on this? I don't follow why we'd want to be able to put this into the datastore. The client only needs to open this and read the head message.

willscott: having one code path for serialization / de-serialization is one less code path to go wrong, but not a strong ask

httpsync/message.go
```diff
@@ -0,0 +1,96 @@
...
+	CidStr string
+}
+
+// SignedHeadMsgEnvelope is the signed envelope of a HeadMsg. It includes the
+// public key of the signer so the receiver can verify it and convert it to a
+// peer id. Note, the receiver is not required to use the provided public key.
+// Also note the signature is over the whole marshalled headmsg instead of just
+// a cidStr. This protects us from forgetting to sign something in the future if
+// we add to HeadMsg.
+type SignedHeadMsgEnvelope struct {
```
MarcoPolo, line 22: I'm thinking we should add json struct tags to rename this to be more json-like.

httpsync/message.go
MarcoPolo: Might be a good exercise. And it could explicitly encode how the bytes are represented (as opposed to the implicit base64 conversion).

MarcoPolo:
@gammazero A couple things changed since I created this. I updated the description. At a high level it uses IPLD schema/bindnode. It also does away with the `HeadMsg` intermediary and just uses the cid.

MarcoPolo:
I'm also working on the example http provider (in rust) and want to check how it feels to use this signed interface over there.

httpsync/publisher.go
```diff
@@ -93,13 +92,20 @@ func (p *publisher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// serve the
 		p.rl.RLock()
 		defer p.rl.RUnlock()
-		out, err := json.Marshal(p.root.String())
+
+		if p.privKey == nil {
+			http.Error(w, "Server not initialized", http.StatusInternalServerError)
+			log.Errorw("Failed to load private key to sign head request")
+			return
+		}
```
gammazero, line 100: I think this check should be moved to `NewPublisher`, so that a publisher cannot even be created without proper initialization.

httpsync/publisher.go
```diff
@@ -93,13 +92,20 @@ func (p *publisher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+		marshalledMsg, err := NewEncodedSignedHead(p.root, p.privKey)
 		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			http.Error(w, "Failed to encode", http.StatusInternalServerError)
 			log.Errorw("Failed to serve root", "err", err)
-		} else {
-			_, _ = w.Write(out)
 		}
```
gammazero, line 106: I think there needs to be a return before this line.

httpsync/message.go
```diff
@@ -0,0 +1,124 @@
...
+// public key of the signer so the receiver can verify it and convert it to a
+// peer id. Note, the receiver is not required to use the provided public key.
+type signedHead struct {
+	Head   cidlink.Link
+	Sig    []byte
+	Pubkey []byte
+}
+
+// NewSignedEnvelope returns a new encoded SignedHead
+func NewEncodedSignedHead(cid cid.Cid, privKey ic.PrivKey) ([]byte, error) {
```
gammazero, line 49: Comment does not match function name.

gammazero, line 49: If this function is not used outside the `httpsync` package then please do not export it.  If this is only done to call from a unit test, then make the test part of this package.

httpsync/message.go
```diff
@@ -0,0 +1,124 @@
...
+	var buf bytes.Buffer
+	err = dagjson.Encode(node.Representation(), &buf)
+	if err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+// OpenSignedHead returns the cid from the encoded signed head given a public key
+func OpenSignedHead(pubKey ic.PubKey, SignedHead io.Reader) (cid.Cid, error) {
```
gammazero, line 75: If this function is not used outside the `httpsync` package then please do not export it.

httpsync/message.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return cid.Undef, err
+	}
+	return openSignedHead(pubKey, *envelop)
+}
+
+// OpenSignedHeadWithIncludedPubKey verifies the signature with the
+// included public key, then returns the public key and cid. The caller can
+// use this public key to derive the signer's peer id.
+func OpenSignedHeadWithIncludedPubKey(SignedHead io.Reader) (ic.PubKey, cid.Cid, error) {
```
gammazero, line 86: If this function is not used outside the `httpsync` package then please do not export it.

gammazero (changes requested):

httpsync/publisher.go
MarcoPolo: good catch!

httpsync/message.go
```diff
@@ -45,8 +45,8 @@ type signedHead struct {
 	Pubkey []byte
 }
 
-// NewSignedEnvelope returns a new encoded SignedHead
-func NewEncodedSignedHead(cid cid.Cid, privKey ic.PrivKey) ([]byte, error) {
+// newEncodedSignedEnvelope returns a new encoded SignedHead
```
gammazero, line 48:
```suggestion
// newEncodedSignedHead returns a new encoded SignedHead
```

httpsync/publisher.go
```diff
@@ -35,6 +35,10 @@ var _ http.Handler = (*publisher)(nil)
 // NewPublisher creates a new http publisher, listening on the specified
 // address.
 func NewPublisher(address string, lsys ipld.LinkSystem, peerID peer.ID, privKey ic.PrivKey) (*publisher, error) {
+	if privKey == nil {
+		return nil, errors.New("private key is nil. Can't sign head requests")
```
gammazero, line 39:
```suggestion
		return nil, errors.New("private key required to sign head requests")
```
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

gammazero (approved):
Looks good with a couple of suggested changes.

httpsync/publisher.go
MarcoPolo: > Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation

I did neither of those things?

PR #68, opened 2022-01-25
No op if we are already synced

MarcoPolo:
## Context

There's a surprising behavior where if you call subscriber.Sync twice with no selector and cid.Undef you will traverse the whole chain instead of stopping on the latestSync.

For example:
Assume publisher `P` has a linked list of blocks:
A <- B <- C.

When we call
```
sub.Sync(ctx, pID, cid.Undef, nil, pAddr)
```
We set `h.latestSync == C`. And we call our block hook fn for `C, B, A`.

If we then call 
```
sub.Sync(ctx, pID, cid.Undef, nil, pAddr) // or C instead of cid.Undef
```
We should expect this to be a no-op since we are already synced. However what actually happens is that we call our block hook fn for `C, B, A` again.

This causes some subtle bugs on the indexer that doesn't expect this behavior (specifically it expects that once a block has been ingested it won't try to be ingested again. Although indexer has some gaurd rails to prevent inconsistency here).

## Proposed Solution
Check if `h.latestSync == cidToSyncTo` in the handler.handle func since we should already have a lock here.

Added a new test that builds a random linked list and tests this behavior.

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -583,13 +583,17 @@ func (h *handler) handleAsync(ctx context.Context, nextCid cid.Cid, ss ipld.Node
 // is locked.
 func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wrapSel, updateLatest bool, syncer Syncer) error {
 	if wrapSel {
-		sel = ExploreRecursiveWithStopNode(h.subscriber.syncRecLimit, sel, h.latestSync)
-
+		// Note this branch is nested under wrapSel because wrapSel adds the
+		// semantics that we stop when we hit the `h.latestSync` node. This is a
+		// special case where we are starting at the stop node so we can just
+		// return.
 		if h.latestSync != nil && h.latestSync.(cidlink.Link).Cid == nextCid {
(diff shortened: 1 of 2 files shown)


subscriber.go
```diff
@@ -582,6 +582,12 @@ func (h *handler) handleAsync(ctx context.Context, nextCid cid.Cid, ss ipld.Node
 func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wrapSel, updateLatest bool, syncer Syncer) error {
 	if wrapSel {
 		sel = ExploreRecursiveWithStopNode(selector.RecursionLimitNone(), sel, h.latestSync)
+
+		if h.latestSync != nil && h.latestSync.(cidlink.Link).Cid == nextCid {
```
gammazero, line 586: Shouldn't this be done whether or not `wrapSel` is true?  If so, then move to top of function, and if not then please comment.

gammazero (changes requested):

gammazero (approved):
LGTM - thank you!

PR #71, opened 2022-01-26
Fix dead handler when two handlers sync the same cid

MarcoPolo:
## Context

Since we are only tracking in-progress syncs with a cid if two handlers start syncing the same cid (from two different providers) one will fail and stall indefinitely.

## Proposed Solution

This adds the peer we are syncing from as part of the key for tracking in progress syncs so handlers can't clobber each other (since there is one handler per peer).

### DIFF
--- a/subscriber_test.go
+++ b/subscriber_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"math/rand"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"testing/quick"
 	"time"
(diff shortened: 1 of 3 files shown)


gammazero (approved):

gammazero:
Need to fix race condition in test first.

subscriber_test.go
```diff
@@ -27,6 +29,98 @@ const (
...
+				}
+			}
+
+			subDS := dssync.MutexWrap(datastore.NewMapDatastore())
+			subLsys := test.MkLinkSystem(subDS)
+			subHost := test.MkTestHost()
+
+			calledTimes := 0
+			sub, err := legs.NewSubscriber(subHost, subDS, subLsys, testTopic, nil, legs.BlockHook(func(i peer.ID, c cid.Cid) {
+				calledTimes++
```
gammazero, line 78: Doesn't this need to be synchronized if the same block hook is called concurrently for different publishers?

subscriber_test.go
```diff
@@ -27,6 +29,98 @@ const (
...
+			}
+
+			wg := sync.WaitGroup{}
+			// Now sync again. We shouldn't call the hook.
+			for _, pub := range publishers {
+				wg.Add(1)
+
+				go func(pub pubMeta) {
+					defer wg.Done()
+					_, err = sub.Sync(context.Background(), pub.h.ID(), cid.Undef, nil, pub.h.Addrs()[0])
```
gammazero, line 91: Multiple goroutines writing to non-local `err` here.

subscriber_test.go
MarcoPolo: Yes thank you

PR #72, opened 2022-01-27
Handle additional datatransfer state

gammazero:
Handle additional datatransfer events that are terminal conditions and will not be followed by additional events.


hannahhoward (commented):
Hey,

Sorry this is confusing. There are multiple paths via which we reach a terminal status in data transfer. Trying to figure it out all through the event can by challenging.

My recommendation here is to not worry about the event, and instead inspect
`channelState.Status()`

There are three terminal statuses:
```	
	datatransfer.Cancelled,
	datatransfer.Completed,
	datatransfer.Failed
```

Only completed is a success. For the other two, they are errors and you can read the error by inspecting `channelState.Message()`. 

That's pretty much all you need to do. I apologize that `go-data-transfer` is a piece of shit library.

PR #73, opened 2022-01-28
Add API to set head CID without publishing it

### LINKED ISSUE #66: Expose an API to set the initial head CID without publishing it
Allow setting initial head CID for legs head publisher. This is useful in a case where legs is used by stateful publisher services that maintain a reference to the latest published head that may be restarted. 
Currently what happens is that the head remains as `cid.Undef` until `UpdateRoot*` is called on the publisher.

Note that we can't use the current `UpdateRoot*` API to do this since that API also publishes a new message onto gossipsub/http depending on the publisher implementation.

Required by https://github.com/filecoin-project/index-provider/issues/145

gammazero:
Fixes #66

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -10,6 +10,8 @@ import (
 
 // Publisher is an interface for updating the published dag.
 type Publisher interface {
+	// SetRoot sets the root CID without publishing it.
+	SetRoot(context.Context, cid.Cid) error
 	// Publishes an update for the DAG in the pubsub channel.
 	UpdateRoot(context.Context, cid.Cid) error
 	// Publishes an update for the DAG in the pubsub channel using custom multiaddrs.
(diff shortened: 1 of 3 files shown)


willscott (approved):

interface.go
```diff
@@ -10,6 +10,8 @@ import (
 
 // Publisher is an interface for updating the published dag.
 type Publisher interface {
+	// SetRoot sets the root CID without publishing it.
+	SetRoot(context.Context, cid.Cid) error
 	// Publishes an update for the DAG in the pubsub channel.
```
masih, line 15:
```suggestion
	// UpdateRoot sets the root CID and publishes its update in the pubsub channel.
```

masih (approved):

PR #75, opened 2022-01-28
Send error if syncer is closed

MarcoPolo:
## Context

If we were closed a subscriber while syncing callers would believe that the sync completed successfully since no error returned. There was no way for a caller to know the sync failed (which it did because the whole thing shutdown)

For example if we're syncing chain A<-B<-C, and we shutdown the subscriber after syncing A but before finishing B, the caller would think that the whole chain was synced successfully.

## Proposed solution

Propagate an error that the sync was closed.

### DIFF
--- a/dtsync/sync.go
+++ b/dtsync/sync.go
@@ -96,6 +96,7 @@ func (s *Sync) Close() error {
 		log.Warnf("Closing datatransfer sync with %d syncs in progress", len(s.syncDoneChans))
 	}
 	for _, ch := range s.syncDoneChans {
+		ch <- errors.New("sync closed")
 		close(ch)
 	}
 	s.syncDoneChans = nil


dtsync/sync.go
```diff
@@ -96,6 +96,7 @@ func (s *Sync) Close() error {
 		log.Warnf("Closing datatransfer sync with %d syncs in progress", len(s.syncDoneChans))
 	}
 	for _, ch := range s.syncDoneChans {
+		ch <- errors.New("sync closed")
```
gammazero, line 99: Should this be done in a `select` in case there are no readers, or do we want to wait for the channel to be read?

MarcoPolo: We don’t want a caller to think a sync succeeded when it didn’t. And anyways this is buffered.

gammazero: Probably safe since the channel is buffered

gammazero (approved):

PR #77, opened 2022-01-29
Do not log error and return same error

gammazero:
Logging an error and then returning that error leads to logging the same errors multiple times.  This decreases the density of useful information and decreases the value of logs.  By returning an error and logging it at a higher level, the error message can contain more meaningful information about the error.

- Log consistently using structured logging
- Update datatransfer and libp2p


dtsync/message.go
```diff
@@ -57,15 +58,13 @@ func DecodeMessage(data []byte) (Message, error) {
 	for len(data) != 0 {
 		val, n, err := varint.FromUvarint(data)
 		if err != nil {
-			log.Errorf("Failed to decode message: cannot read number of multiadds: %s", err)
-			return Message{}, err
+			return Message{}, fmt.Errorf("cannot read number of multiadds: %s", err)
```
willscott, line 61: prefer '%w' on Errorf of errors so that the error can be unwrapped to get the internal error. here and below

gammazero: Done

willscott (approved):

PR #79, opened 2022-02-02
ScopedBlock hook with subscriber.SyncWithHook

MarcoPolo:
Adds the ability to specify a block hook in a `Sync` call, and have that block hook be called for blocks associated with that sync. Could help clean up logic in indexer and usages of block hook in general. 

For example instead of having to keep track of a mapping of entriesCid -> AdCid like we currently do, we could rely on closured variables to get that information. For example:

```go
// in syncAdEntries in linksystem.go in indexer
      
	_, err = ing.sub.Sync(ctx, from, entriesCid, ing.entriesSel, nil, func(_ peer.ID, c cid.Cid) {
		node, err := decodeIPLDNode(c.Prefix().Codec, bytes.NewBuffer(val))
	        if err != nil {
		        log.Errorw("Error decoding ipldNode", "err", err)
		        return
	        }
	    	err = ing.indexContentBlock(adCid, from, node)
	          if err != nil {
		          log.Errorw("Error processing entries for advertisement", "err", err)
	          } else {
		          log.Info("Done indexing content in entry block")
		          ing.signalMetricsUpdate()
	          }
       })
```

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -72,6 +72,11 @@ type Subscriber struct {
 	handlers      map[peer.ID]*handler
 	handlersMutex sync.Mutex
 
+	// A map of block hooks to call for a specific peer id, instead of the general
+	// block hook func.
+	scopedBlockHook      map[peer.ID]BlockHookFunc
+	scopedBlockHookMutex *sync.RWMutex
+
 	// inEvents is used to send a SyncFinished from a peer handler to the
 	// distributeEvents goroutine.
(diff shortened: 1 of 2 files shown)


gammazero (commented):
This looks good so far (one minor comment).

gammazero:
Can this be closed because #89 includes it?

MarcoPolo:
Closing because #89 brings it in

PR #80, opened 2022-02-08
Add optional miner ID to pubsub message

gammazero:
By carrying the miner ID in the pubsub message the lotus chain-nodes that relay the messages can lookup the merket process peer ID and ensure that was the peer that signed the message.  This authenticates the message as being sent from a peer that is known on-chain.

### DIFF
--- a/dtsync/option.go
+++ b/dtsync/option.go
@@ -24,8 +24,8 @@ func (c *config) apply(opts []Option) error {
 	return nil
 }
 
-// MinerID sets a miner ID to include in the pubsub message.
-func MinerID(minerID string) Option {
+// WithMinerID sets a miner ID to include in the pubsub message.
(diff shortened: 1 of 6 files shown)


dtsync/option.go
```diff
@@ -23,6 +24,14 @@ func (c *config) apply(opts []Option) error {
 	return nil
 }
 
+// MinerID sets a miner ID to include in the pubsub message.
+func MinerID(minerID string) Option {
```
willscott, line 28: this can be general at this interface too: `WithExtraData(data []byte)`

willscott (approved):

dtsync/option.go
masih: I too think we need to keep this Filecoin agnostic.

willscott (approved):

PR #82, opened 2022-02-09
Fix blockHook semantics in httpsync subscriber

MarcoPolo:
## Context

graphsyncs block hooks get called even if the block is received from the local datastore. We rely on this property in storetheindex's ingester. Previously the httpsync implementation only called the block hook when receiving the block from the remote http endpoint.

What made this problem especially tricky is that the indexer deletes the block from the datastore after processing. And that happens before we continue traversing in go-legs subscriber. So what ended up happening is that we deleted the block after we fetched it from the remote but *before* we continued our traversal. This meant that go-legs didn't know how to traverse from there.

## Proposed solution

Track the traversed block CIDs during traversal. Then at the end call the block hook function. More details in the comment in the code.

Also added some more tests and fixed an incorrect test.


willscott (approved):

willscott:
Thanks for tracking this down!

gammazero (approved):

PR #85, opened 2022-02-11
HTTP Peerstore

MarcoPolo:
## Context

If a user called `.Sync` on on a peer with a nil address we would always try to use graphsync on it. This is a problem if the peer is really an http provider so it's not reachable via graphsync.

Internally graphsync would check the libp2p host's peerstore to get a recently used address for that peer.

## Proposed solution

Create a new http only peerstore, and remember peers' http address when we first get them.

* Why not re-use the libp2p host's peerstore?
  * We have no guarantee what other users of the host will do with this address. They may use it as an endpoint to connect to via libp2p, and that will fail.
  * The peerstore will broadcast address changes. We don't really want this for the same reason as above.
  * An http-only user shouldn't have to create a libp2p host just to get this functionality.
  
Currently no error handling and culling of the addresses (besides TTL) happens. 
 
## Semantics

* If a nil peerAddr is passed in and that peer has both an http and libp2p endpoint, the implementation will prefer the http endpoint.

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -367,9 +367,7 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid, 
 				break
 			}
 		}
-	}
-
-	if peerAddr == nil {
+	} else {
 		// Check if we have an http url for this peer since we didn't get a peerAddr.
 		// Note that this gives a preference to use httpSync over dtsync if we have
 		// seen http address and we called sync with no explicit peerAddr.
--- a/subscriber.go
+++ b/subscriber.go
@@ -358,12 +358,10 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid, 
 	var syncer Syncer
 
 	isHttpPeerAddr := false
-	shouldAddToHttpPeerstore := false
 	if peerAddr != nil {
 		for _, p := range peerAddr.Protocols() {
 			if p.Code == multiaddr.P_HTTP || p.Code == multiaddr.P_HTTPS {
 				isHttpPeerAddr = true
-				shouldAddToHttpPeerstore = true
 				break
 			}
 		}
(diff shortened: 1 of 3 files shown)


MarcoPolo:
I think it should wait until it successfully syncs before it adds the address to the peerstore.

subscriber.go
```diff
@@ -345,17 +357,37 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid,
...
-					return cid.Undef, fmt.Errorf("cannot create http sync handler: %w", err)
-				}
+				isHttpPeerAddr = true
+				shouldAddToHttpPeerstore = true
 				break
 			}
 		}
+	}
+
+	if peerAddr == nil {
```
gammazero, line 372: Looks like this can be an `else`

subscriber.go
```diff
@@ -345,17 +357,37 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid,
...
+		// Check if we have an http url for this peer since we didn't get a peerAddr.
+		// Note that this gives a preference to use httpSync over dtsync if we have
+		// seen http address and we called sync with no explicit peerAddr.
+		possibleAddrs := s.httpPeerstore.Addrs(peerID)
```
gammazero, line 376: If we have both addresses, do we want to prefer one over the other?  Or, should we discard the older one?  Or, should we try to sync with one, and if that fails try the other?

If a publisher ever uses an http address, it means they will be stuck using that, even if they later have a libp2p address.  That might be a reason to discard the older address.

subscriber.go
```diff
@@ -423,6 +455,12 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid,
 		return cid.Undef, fmt.Errorf("sync handler failed: %w", err)
 	}
 
+	if shouldAddToHttpPeerstore {
+		// Store this http address so that future calls to sync will work without a
+		// peerAddr (given that it happens within the TTL)
+		s.httpPeerstore.AddAddr(peerID, peerAddr, peerstore.ProviderAddrTTL)
```
gammazero, line 461: 💯

gammazero (approved):
Looks good, with a couple of comments.  This is great because now when we automatically poll publishers it will work for both http and libp2p addresses.

subscriber.go
MarcoPolo: That has a preference for HTTP for no specific reason. If a publisher ever uses an http address it will keep trying to use that until the addresses' TTL expires (currently 30 minutes).

I think it's okay to not try to use libp2p as a fallback unless we see a strong need for it. But open to discussion.

MarcoPolo:
A couple changes I think we want:
1. Call `AddAddr` everytime we reconnect to that peer with the peerAddr so that we extend the TTL.
2. use `s.addrTTL` instead of provideraddrttl
3. Just like how we are waiting until the end to call AddAddr in the http peerstore, we should do the same for the libp2p peerstore store since we don't know if it's valid until after we connect.

MarcoPolo:
@gammazero updated with the changes from my comment above

subscriber.go
MarcoPolo: to expand a bit:

1. I think it's unlikely a publisher will have both a libp2p publisher and http publisher for the same peer ID.
2. In the case that they do, it's probably safe to prefer the http publisher since it's less resource intensive for the publisher (an http cache proxy is much easier than a graphsync cache proxy).
3. If the http endpoint fails we eventually remove it from the peerstore via the TTL mechanism.

gammazero (approved):
LGTM

PR #89, opened 2022-02-16
SyncFinished Event now includes CIDs traversed

MarcoPolo:
We keep track of the cids traversed during a sync so that the syncfinished event has that information. This would remove the hack around cidwaiter and maybe simplify the implementation of the ingester proposal.

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -186,17 +186,7 @@ func NewSubscriber(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem, 
 	scopedBlockHookMutex, scopedBlockHook, blockHook := wrapBlockHook(cfg.blockHook)
 
 	var dtSync *dtsync.Sync
-	if cfg.dtManager != nil {
-		if ds != nil {
-			log.Warn("Datastore cannot be used with DtManager option")
-		}
-		if cfg.blockHook != nil {
-			log.Warn("BlockHook option cannot be used with DtManager option")
-		}
-		dtSync, err = dtsync.NewSyncWithDT(host, cfg.dtManager)
-	} else {
-		dtSync, err = dtsync.NewSync(host, ds, lsys, blockHook)
-	}
+	dtSync, err = dtsync.NewSync(host, ds, lsys, blockHook)
 	if err != nil {
 		cancelPubsub()
 		return nil, err
--- a/subscriber.go
+++ b/subscriber.go
@@ -114,6 +114,17 @@ type SyncFinished struct {
 	Cid cid.Cid
 	// PeerID identifies the peer this SyncFinished event pertains to.
 	PeerID peer.ID
+	// A list of cids that this sync acquired. In order from latest to oldest. The latest cid will always be at the beginning.
+	SyncedCids []cid.Cid
+}
+
+func WrapBlockHookWithSyncedCidTracker(cidsSeenSoFar *[]cid.Cid, blockHook BlockHookFunc) BlockHookFunc {
+	return func(p peer.ID, c cid.Cid) {
+		*cidsSeenSoFar = append(*cidsSeenSoFar, c)
+		if blockHook != nil {
+			blockHook(p, c)
+		}
+	}
 }
 
 // handler holds state that is specific to a peer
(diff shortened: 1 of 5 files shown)


subscriber.go
```diff
@@ -626,9 +669,24 @@ func (h *handler) handleAsync(ctx context.Context, nextCid cid.Cid, ss ipld.Node
...
+	syncedCids := []cid.Cid{}
+	hook = WrapBlockHookWithSyncedCidTracker(&syncedCids, hook)
+	scopedBlockHookMutex.Lock()
+	scopedBlockHook[h.peerID] = hook
+	scopedBlockHookMutex.Unlock()
+	defer func() {
+		scopedBlockHookMutex.Lock()
+		if scopedBlockHook == nil {
+			scopedBlockHook = make(map[peer.ID]BlockHookFunc)
+		}
```
gammazero, line 685: Looks unneeded.  It is OK to delete from nil map.

subscriber.go
```diff
@@ -626,9 +669,24 @@ func (h *handler) handleAsync(ctx context.Context, nextCid cid.Cid, ss ipld.Node
...
-func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wrapSel, updateLatest bool, syncer Syncer) error {
+func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wrapSel, updateLatest bool, syncer Syncer, hook BlockHookFunc, scopedBlockHookMutex *sync.RWMutex, scopedBlockHook map[peer.ID]BlockHookFunc) error {
 	log := log.With("cid", nextCid, "peer", h.peerID)
 
+	// This is not set to nil so we can get a pointer.
+	syncedCids := []cid.Cid{}
+	hook = WrapBlockHookWithSyncedCidTracker(&syncedCids, hook)
+	scopedBlockHookMutex.Lock()
+	scopedBlockHook[h.peerID] = hook
+	scopedBlockHookMutex.Unlock()
```
gammazero, line 680: Change these to `h.subscriber.scopedBlockHookMutex` and `h.subscriber.scopedBlockHook`

subscriber.go
```diff
@@ -663,7 +721,7 @@ func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wr
 
 	// Tell the Subscriber to distribute SyncFinished to all notification
 	// destinations.
-	h.subscriber.inEvents <- SyncFinished{Cid: nextCid, PeerID: h.peerID}
+	h.subscriber.inEvents <- SyncFinished{Cid: nextCid, PeerID: h.peerID, SyncedCids: syncedCids}
```
gammazero, line 724: Do we need both `Cid` and `SyncedCids`? Or will the `nextCid` always be the first item in `syncedCids`?

gammazero (commented):
See comments

If this can be used to replace the cidWaiter and reverse entry-ad mapping in storetheindex, this would be great.

subscriber.go
```diff
@@ -444,7 +487,7 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid,
 	hnd.syncMutex.Lock()
 	defer hnd.syncMutex.Unlock()
 
-	err = hnd.handle(ctx, nextCid, sel, wrapSel, updateLatest, syncer)
+	err = hnd.handle(ctx, nextCid, sel, wrapSel, updateLatest, syncer, hook, s.scopedBlockHookMutex, s.scopedBlockHook)
```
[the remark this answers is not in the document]
MarcoPolo: Ah I was thinking of putting a reference to subscriber in the handler, glad that it was already there.

subscriber.go
MarcoPolo: `nextCid` will always be the first item in syncedCids. I didn't want to change the api, but I think we can do this.

MarcoPolo: We can do this. It's just a bigger breaking change. Maybe in another PR? What do you think?

gammazero: That's fine

gammazero (approved):
LGTM

MarcoPolo:
Ran test that failed locally with the failed inputs and it passes, so assuming that is a flake for now.

PR #91, opened 2022-02-18
Allow head query over libp2p using dns address

### LINKED ISSUE #221: Ingest/Announce doesn't work when multiaddr has a dnsname
tried to ingest with the multiaddr of the form
`/dns4/<dnsname>/tcp/4567/p2p/<id>` and the ingest fails with the error:
`cannot query head for sync: Get \"http://unused.invalid/head\": no good addresses`

When an ip is resolved before announce, this error does not occur.

gammazero:
Need to use `host.Connect` to  resolve DNS multiaddresses.

Fixes https://github.com/filecoin-project/storetheindex/issues/221

### DIFF
--- a/p2p/protocol/head/head.go
+++ b/p2p/protocol/head/head.go
@@ -62,11 +62,7 @@ func QueryRootCid(ctx context.Context, host host.Host, topic string, peerID peer
 				if err != nil {
 					return nil, err
 				}
-				conn, err := gostream.Dial(ctx, host, peerID, deriveProtocolID(topic))
-				if err != nil {
-					return nil, err
-				}
-				return conn, nil
+				return gostream.Dial(ctx, host, peerID, deriveProtocolID(topic))
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


p2p/protocol/head/head.go
```diff
@@ -51,11 +51,22 @@ func (p *Publisher) Serve(host host.Host, topic string) error {
...
 			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
-				return gostream.Dial(ctx, host, peer, deriveProtocolID(topic))
+				addrInfo := peer.AddrInfo{
+					ID: peerID,
+				}
+				err := host.Connect(ctx, addrInfo)
+				if err != nil {
+					return nil, err
+				}
+				conn, err := gostream.Dial(ctx, host, peerID, deriveProtocolID(topic))
```
willscott, line 65: unfortunate!

willscott (approved):

Issue #95, opened 2022-02-23
legs subscriber must allow datatrasfer to be optionally set


masih:
Otherwise, a lib2p host with go-legs subscriber cannot start a graphsync server, since registration of stream handler for graphsync server is singular.

It seems limiting to require users of go-legs to not start up a graphsync server if they are running a go-legs subscriber.

Consider adding the option back which was removed [here](https://github.com/filecoin-project/go-legs/pull/89/files#diff-086a37ef11c8fd9a2516991339db79c207741ac98dd89c8215d31a8bead2814fL64-L71).

Consider alternative approaches to guaranteeing the correctness of blockhook mechanism.

masih:
Changing that option to take both `datatransfer.Manager` and `graphsync.GraphExchange` should be sufficient to allow both registration of block hook at root and instantiate if not set.

willscott:
or we can add it back on just on the publisher side, right?

masih:
It was never removed from the publisher side.

MarcoPolo:
Done in #96

PR #96, opened 2022-02-23
Re-enable DtManager subscriber option

gammazero:
Without this option, if a process uses a go-legs subscriber, it cannot expose a graphsync instance on the same libp2p host.

There can only be a single stream handler for a protocol ID and graphsync instantiation sets it for "/ipfs/graphsync/1.0.0" this protocol ID.  So we need this option back?

### DIFF
--- a/dtsync/sync.go
+++ b/dtsync/sync.go
@@ -39,11 +39,15 @@ type Sync struct {
 
 // NewSyncWithDT creates a new Sync with a datatransfer.Manager provided by the
 // caller.
-func NewSyncWithDT(host host.Host, dtManager dt.Manager) (*Sync, error) {
+func NewSyncWithDT(host host.Host, dtManager dt.Manager, gs graphsync.GraphExchange, blockHook func(peer.ID, cid.Cid)) (*Sync, error) {
 	registerVoucher(dtManager)
 	s := &Sync{
 		host:      host,
 		dtManager: dtManager,
+	}
+
+	if blockHook != nil {
+		s.unregHook = gs.RegisterIncomingBlockHook(makeIncomingBlockHook(blockHook))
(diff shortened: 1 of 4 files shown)


willscott:
We need this to work on the publisher side primarily, right?

gammazero:
@willscott This is a subscriber option.  I believe @masih has more detail, and hopefully @MarcoPolo can weigh in on whether this is usable.

masih:
> We need this to work on the publisher side primarily, right?

On publisher side, it was never removed. dtpublisher has a dedicated constructor for supplying dt and does not use options.

MarcoPolo:
We need to figure out how to add block hooks here, otherwise the indexer will be broken. 

Just so I know, why do we need to re enable this?

masih:
> Why do we need to re enable this?

From a general usability perspective, go-legs aims to be a generic protocol for syncing DAGs. 
A `Subscriber` instantiating its own graphsync means that on other instances can be instantiated on the same libp2p host.
As a generic protocol, it seems odd to enforce the limitation that if a legs subscriber is instantiated on the same libp2p host, that host cannot run a graphsync server. 

In terms of concrete use cases today, there are tests in index-provider that used to do this. There is also Pando, which uses go-legs to subscribe to publishers of metadata and makes them available for retrieval among doing other things.

Also see #95

MarcoPolo:
Now you need to pass in the graphsync.graphexchange when using the dtmanager option. This way we can register our incoming block hooks still.

MarcoPolo (approved):
I'm approving since I'm not sure if it'll let @gammazero approve their own PR. but please tal too @gammazero

gammazero (commented):
LGMT

PR #97, opened 2022-02-23
Generalize CheckAlreadySynced

MarcoPolo:
## Context
When the stop node in the recursive selector is the same as our sync node, we will never stop early because the stop node is only checked after traversal has started. 

We previously did a check to see if our sync cid was our latestSync and exit early. This is a special case of the more general form where the stop node in the selector is the cid of the node we are going to start traversal on.

Then we added a special flag to always do this special case.

## Proposed Solution

Instead of having the user remember to pass in the flag for the special case check, lets try to determine if we're already in the case where we are trying to stop on the same cid as where we are starting.

Note that this uses knowledge of the Recursive Selector structure to get the stop node. There's a test that will break if this knowledge is ever wrong. It's not ideal, but I think it's better than having to remember to pass in the special flag.

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -478,7 +478,7 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid, 
 	hnd.syncMutex.Lock()
 	defer hnd.syncMutex.Unlock()
 
-	err = hnd.handle(ctx, nextCid, sel, wrapSel, updateLatest, syncer, cfg.scopedBlockHook, cfg.checkAlreadySynced)
+	err = hnd.handle(ctx, nextCid, sel, wrapSel, updateLatest, syncer, cfg.scopedBlockHook)
 	if err != nil {
 		return cid.Undef, fmt.Errorf("sync handler failed: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


subscriber.go
```diff
@@ -676,24 +676,13 @@ func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wr
...
 		sel = ExploreRecursiveWithStopNode(h.subscriber.syncRecLimit, sel, h.latestSync)
-	} else if checkAlreadySynced {
-		// Special case when starting at stop node; selector does not check it.
-		if h.latestSync != nil && h.latestSync.(cidlink.Link).Cid == nextCid {
-			// Nothing to do. We've already synced to this cid because we have
-			// it as h.latestSync.
-			log.Infow("Already synced")
+	}
+
+	if sel != nil {
```
gammazero, line 682: How can `sel` ever be `nil` here?.  If `wrapSel` is `true`, then `sel` gets assigned.  If `wrapSel` is false, then `sel` was already non-nil.

subscriber.go
```diff
@@ -676,29 +676,16 @@ func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wr
...
-		if h.latestSync != nil && h.latestSync.(cidlink.Link).Cid == nextCid {
-			// Nothing to do. We've already synced to this cid because we have
-			// it as h.latestSync.
-			log.Infow("Already synced")
-			return nil
-		}
 	}
 
-	err := syncer.Sync(ctx, nextCid, sel)
+	stopNode, err := getStopNode(sel)
```
gammazero, line 682: The `err` is completely discarded.  If is really is an error, then log it.  If it is only used to determine if a stopNode was found, then have `getStopNode` return a bool instead.

gammazero (approved):

gammazero:
Please bump the version and I will refer to that in https://github.com/filecoin-project/storetheindex/pull/234

PR #98, opened 2022-02-24
Change temporary address TTL

gammazero:
Once there is a stream reset, there will not be any more datatransfer events, so it must be streated as a terminal state.

Use a much longer TTL for temporary addresses, so they remain in the peersotre until the (very long) ad chain is completely synced.

### DIFF
--- a/dtsync/sync.go
+++ b/dtsync/sync.go
@@ -176,12 +176,6 @@ func (s *Sync) onEvent(event dt.Event, channelState dt.ChannelState) {
 		if strings.HasSuffix(msg, "content not found") {
 			err = errors.New(err.Error() + ": content not found")
 		}
-	case dt.Ongoing:
-		if channelState.Message() != "stream reset" {
-			return
-		}
-		log.Warnw("Detected stream reset from peer", "peer", channelState.OtherPeer())
-		err = errors.New("stream reset")
 	default:
 		// Ignore non-terminal channel states.
 		return
--- a/dtsync/sync.go
+++ b/dtsync/sync.go
@@ -176,6 +176,12 @@ func (s *Sync) onEvent(event dt.Event, channelState dt.ChannelState) {
 		if strings.HasSuffix(msg, "content not found") {
 			err = errors.New(err.Error() + ": content not found")
 		}
+	case dt.Ongoing:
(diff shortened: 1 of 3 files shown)


dtsync/sync.go
```diff
@@ -176,6 +176,12 @@ func (s *Sync) onEvent(event dt.Event, channelState dt.ChannelState) {
 		if strings.HasSuffix(msg, "content not found") {
 			err = errors.New(err.Error() + ": content not found")
 		}
+	case dt.Ongoing:
+		if channelState.Message() != "stream reset" {
```
willscott, line 180: will having this be an error state mean the next layer up re-starts / re-attempts sync?

willscott (approved):

dtsync/sync.go
gammazero: I don't know and I need @hannahhoward  to look at it.  As far as I can tell, once that is seen there is no more activity for that graphsync/datatransfer session.

MarcoPolo: I believe this will cause the sync call to fail, so it's up to the caller to do a resync. For the indexer, we would do this next time we do a sync.

MarcoPolo:
Should this be fixed in datastransfer instead?

willscott:
yes - @hannahhoward says there's a restart manager that should be configured in the go-datatransfer layer, and probably isn't right now

dtsync/sync.go
hannahhoward: it means the connection between the remote peer and the local peer got disconnected. It most cases the data transfer should attempt to restart the connection to get things going again, though you have to configure the channel monitor that does the restarts. When no restarts are configured one one expect it to become a failure but I didn't right the channel monitor and it sounds like it doesn't put the channel in a failure state. The SPX data transfer instance is configured for no restarts, so I would configure it on Store The Index --- you can see an example of how to do so here: https://github.com/filecoin-project/lotus/blob/master/node/modules/client.go#L120 or here -- https://github.com/application-research/filclient/blob/master/filclient.go#L186

MarcoPolo (approved):
Note that the PR title is no longer accurate. This just changes the temp addr ttl.

PR #102, opened 2022-03-01
Fix possible deadlock if user places backpressure

MarcoPolo:
## Context
If a user placed backpressure by not pulling on the OnSyncFinished eventchannel then we would prevent any handler from running for thatpublisher.

This could lead to a deadlock like the one in storetheindex. Where the indexer applies backpressure if the ingest workers cannot keep up. The ingest workers were deadlocked because they could not finish their work because it required another call to `subscriber.Sync` to fetch the ad entries. But the sync was blocked on a handler that was blocked on sending to a channel that was blocked by the indexer pulling the channel that was blocked by not having a worker available to send work to that was blocked by the worker not being to make progress...

## Proposed change

Introduce a new lock that protects around setting the latestSync and sending updates SyncFinished events. And limit the scope of the syncMutex to _only_ the part that needs to actually do the syncing. This means that even if legs is blocked by backpressure, it can still do a specific selector/non-updateLatest sync if requested. In the indexer use-case this allows workers to make progress while still being able to apply back-pressure.

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -136,10 +136,9 @@ type handler struct {
 	// syncMutex serializes the handling of individual syncs. This should only
 	// guard the actual handling of a sync, nothing else.
 	syncMutex sync.Mutex
-	// When grabbing this lock, make sure to encompass the syncMutex. It's
-	// incorrect to grab this lock only while updating latestSync. You must hold
-	// this lock while syncing to the next latestSync. Otherwise another process
-	// may overwrite your state.
+	// If updating this from the result of a sync make sure you grab this lock
+	// before calling handler.handle(). Otherwise another process may overwrite
+	// your state.
 	latestSyncMu sync.Mutex
 	latestSync   ipld.Link
 	msgChan      chan cid.Cid
--- a/subscriber.go
+++ b/subscriber.go
@@ -499,9 +499,7 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid, 
 		defer hnd.latestSyncMu.Unlock()
 	}
 
-	hnd.syncMutex.Lock()
 	syncedCids, err := hnd.handle(ctx, nextCid, sel, wrapSel, syncer, cfg.scopedBlockHook)
-	hnd.syncMutex.Unlock()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


willscott (approved):

MarcoPolo:
I cleaned up a bit of the code and added a test. Code I get another review @willscott ?

subscriber.go
```diff
@@ -133,12 +133,18 @@ func WrapBlockHookWithSyncedCidTracker(cidsSeenSoFar *[]cid.Cid, blockHook Block
 // handler holds state that is specific to a peer
 type handler struct {
 	subscriber *Subscriber
-	latestSync ipld.Link
-	msgChan    chan cid.Cid
+	// syncMutex serializes the handling of individual syncs. This should only
+	// guard the actual handling of a sync, nothing else.
+	syncMutex sync.Mutex
+	// When grabbing this lock, make sure to encompass the syncMutex. It's
```
willscott, line 139: maybe worth being super explicit here:
`this lock must be claimed before syncMutex to deadlocks`

subscriber.go
```diff
@@ -486,26 +492,38 @@ func (s *Subscriber) Sync(ctx context.Context, peerID peer.ID, nextCid cid.Cid,
...
-	err = hnd.handle(ctx, nextCid, sel, wrapSel, updateLatest, syncer, cfg.scopedBlockHook)
+	syncedCids, err := hnd.handle(ctx, nextCid, sel, wrapSel, syncer, cfg.scopedBlockHook)
 	if err != nil {
 		return cid.Undef, fmt.Errorf("sync handler failed: %w", err)
 	}
 
+	if updateLatest {
+		hnd.latestSync = cidlink.Link{Cid: nextCid}
+		hnd.subscriber.inEvents <- SyncFinished{Cid: nextCid, PeerID: hnd.peerID, SyncedCids: syncedCids}
+		log.Infow("Updating latest sync")
```
willscott, line 510: make this log more useful: include peer, cid

willscott (approved):

subscriber.go
gammazero: That is already done by `log.With` above.

PR #104, opened 2022-03-04
Specify alternate recursion limit for trusted peers

gammazero:
If a peer is trusted then it will use a different recursion limit than an untrusted host. A function to determine if the host is trusted, and the trusted resursion limit can be specified using a subscriber option.


gammazero:
Opting to remove recursion limits and do rate limiting instead

PR #116, opened 2022-04-14
Add ability to get subscriber http peerstore

gammazero:
This is needed for getting the HTTP publisher address when an advertisement is retreived via HTTP.


http_test.go
```diff
@@ -167,7 +167,7 @@ func TestSyncFnHttp(t *testing.T) {
 	cids, _ := test.RandomCids(1)
 	ctx, syncncl := context.WithTimeout(context.Background(), time.Second)
 	defer syncncl()
-	syncCid, err := te.sub.Sync(ctx, te.srcHost.ID(), cids[0], nil, te.pubAddr)
+	_, err := te.sub.Sync(ctx, te.srcHost.ID(), cids[0], nil, te.pubAddr)
 	if err == nil {
```
willscott, line 171:
```suggestion
	if _, err := te.sub.Sync(ctx, te.srcHost.ID(), cids[0], nil, te.pubAddr); err == nil {
```

willscott (approved):

PR #119, opened 2022-04-21
Do not let async goroutines build up

gammazero:
When announce pubsub messages arrive for a peer, and the Subscriber is already syncing that peer, these message will create goroutines that are waiting to be dismissed when the Subscriber is done.  If there are a large number of announce messages received during an ongoing sync, this uses significant memory.  This fix changes the way a new pending CID replaces the previous pending CID in order to never allow more than one pending goroutine waiting on a pending sync.

* New Subscriber.Announce function

When announce messages are sent by means other than pubsub, to an application using go-legs (such as indexer), the application needs to be able to handle the announce messages in the same manner as with announces received over pubsub.  This means using the pending CID replacement and single pending goroutine logic so that the system is not overloaded with a large number of announce messages.  Previously, an application had to call `Subscriber.Sync` which would cause all these calls to back up waiting for previous Sync calls to complete.  Not the application can call Subscriber.Announce()

* Support HTTP addresses in announce messages over pubsub

The Subscriber will now recognize and handle HTTP addreses from an announce message recevied over pubsub.  This means that a pubsub announce can cause the indexer to sync over HTTP if the message contains an HTTP address. Previously, announce pubsub messages could only be used to tell the Subscriber to sync using datatransfer/graphsync.


willscott (approved):
The test that would be really great to see would be:
* publisher announces
* subscriber begins sync
* publisher blocks on graphsync response so that subscriber doesn't finish
* publisher announces 2 more times
* publisher resumes graphsync responses

validate:
* current sync completes
* of the 2 additional, the newer one pre-empts the older.

willscott (approved):

PR #120, opened 2022-04-26
Fix derived protocol ID so there are no double slashes

### LINKED ISSUE #86: head protocol ID ends up with double `/` for topic names starting with `/`
Example:
```
/legs/head//indexer/ingest/mainnet/0.0.1
```

https://github.com/filecoin-project/go-legs/blob/36b1633ecf46df82742c188a97f21773cb3872fc/p2p/protocol/head/head.go#L40

gammazero:
Fixes #86

### DIFF
--- a/p2p/protocol/head/head.go
+++ b/p2p/protocol/head/head.go
@@ -38,7 +38,7 @@ func NewPublisher() *Publisher {
 }
 
 func deriveProtocolID(topic string) protocol.ID {
-	return protocol.ID("/legs/head/" + topic + "/0.0.1")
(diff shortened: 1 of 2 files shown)


masih (commented):
@gammazero would this be a breaking change? in that, versions that are not upgraded to the latest will not be able to communicate with the newer version to lookup `head`. If so it would be a deal breaker. 

Ideally both should work and we should eventually retire the double slash by bumping the version in protocol ID?

gammazero:
@masih Good catch.  It does not cause indexing, when triggered by an annoiunce to fail, but it does cause explicit sync to fail.  I will revert.

PR #124, opened 2022-05-14
Fix rate limit burst refresh for graphsync

gammazero:
After dt/gs session hits the burst limit, it is necessary wait for the token bucket to refil before restarting the session. The wait was happening but was also emptying its bucket.  So when the session restarted there was no burst and rate limiting would happen immidiately after reading the next block or two.  This fixes this to wait without emptying the bucket.

Other changes:
- Improved logging
- AllowPeerFunc only retruns bool. It is up to the function to log any errors, not legs.  Also returning the error does not allow any decision beyond what the boolean return value already does.

### DIFF
--- a/dtsync/syncer.go
+++ b/dtsync/syncer.go
@@ -59,20 +59,27 @@ func (s *Syncer) Sync(ctx context.Context, nextCid cid.Cid, sel ipld.Node) error
 		// Wait for transfer finished signal.
 		err = <-syncDone
 		if err, ok := err.(rateLimitErr); ok {
-			log.Infow("hit rate limit. Waiting and will retry later", "cid", nextCid, "source_peer", s.peerID)
-
 			// Wait until the rate limit bucket is fully refilled since this is
 			// a relatively heavy operation (essentially restarting the sync).
 			// Note, cannot use s.rateLimiter.WaitN here because that waits,
 			// but also consumes n tokens.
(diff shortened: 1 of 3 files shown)


dtsync/syncer.go
```diff
@@ -60,12 +61,19 @@ func (s *Syncer) Sync(ctx context.Context, nextCid cid.Cid, sel ipld.Node) error
...
-			// Wait until we've fully refilled our rate limit bucket since this is a
-			// relatively heavy operation (essentially restarting the sync).
-			waitErr := s.rateLimiter.WaitN(ctx, s.rateLimiter.Burst())
-			if waitErr != nil {
-				return err
+			// Wait until the rate limit bucket is fully refilled since this is
+			// a relatively heavy operation (essentially restarting the sync).
+			// Note, cannot use s.rateLimiter.WaitN here because that waits,
+			// but also consumes n tokens.
+			waitTime := time.Second * time.Duration(s.rateLimiter.Burst()) / time.Duration(s.rateLimiter.Limit())
```
willscott, line 68: what length of time is this going to be in reality?
while heavy, i wouldn't really imagine waiting more than about 1 second per resumption

willscott (approved):

dtsync/syncer.go
gammazero: Only a second or two to restart sync depending on indexer load.  Well under a second in testing.  If the burst size if 500, and the rate is 100 per second, then the delay will be 5 seconds.

PR #126, opened 2022-05-19
Implement the ability to sync a DAG in segments of configurable depth

masih:
Syncing very large DAGs in a single sync call, either via `httpsync` or
`dtsync` can result in excessive memory usage doe to internal state
stored by underlying sync mechanism. For example, `httpsync` keeps the
traversed path in memory, which can result in high memory usage when the
syncing excessively large dags.

To avoid such issues, break up the syncing of a DAG into a series of
individual syncs with a smaller and configurable depth limit. This would
then provide an opportunity for any ephemeral state stored in memory
within each sync to be garbage collected, reducing the overall memory
footprint when syncing large DAGs.

The changes introduced are:
* An option to control the maximum depth limit for each sync segment.
* an extension to the block hook that lets the user control the flow of
  segmented sync cycles.
* scoped block hook overrides the general block hook if both are set.

For backward compatibility, segmented sync is disabled by default. The
segmented sync is only enabled if:
* the segment depth limit is not a negative value, and
* a block hook, either general or scoped, is set.

The block hook is needed because in go-legs level, we do not have enough
information to decode a synced block and determine what its "next" link
may be. Therefore, the user of the library is required to do this and
control the flow via the segment block hook actions.

The scoped block hook overrides the general block hook, because, the
"actions" extension to the block hook modifies the state of sync.
Running both general and scoped block hook could then make conflicting
calls to the action and result in unexpected behaviour. To avoid such
edge cases, the scoped hook simply overrides the general hook if both
are set. The tests are adjusted to assert this behaviour.

Relates to: https://github.com/filecoin-project/storetheindex/issues/487

### DIFF
--- a/option.go
+++ b/option.go
@@ -34,6 +34,8 @@ type config struct {
 	latestSyncHandler LatestSyncHandler
 
 	rateLimiterFor RateLimiterFor
+
+	segDepthLimit int64
 }
 
 type Option func(*config) error
@@ -91,10 +93,25 @@ func HttpClient(client *http.Client) Option {
 	}
 }
 
-// BlockHook adds a hook that runs when a block is received.
+// BlockHook adds a hook that is run when a block is received via Subscriber.Sync along with a
+// SegmentSyncActions to control the sync flow if segmented sync is enabled.
+// Note that if segmented sync is disabled, calls on SegmentSyncActions will have no effect.
+// See: SegmentSyncActions, SegmentDepthLimit, ScopedBlockHook.
 func BlockHook(blockHook BlockHookFunc) Option {
 	return func(c *config) error {
 		c.blockHook = blockHook
+		return nil
+	}
+}
+
+// SegmentDepthLimit sets the maximum recursion depth limit for a segmented sync.
+// Setting the depth to a value less than zero disables segmented sync completely.
+// Disabled by default.
+// Note that for segmented sync to function at least one of BlockHook or ScopedBlockHook must be
+// set.
+func SegmentDepthLimit(depth int64) Option {
+	return func(c *config) error {
+		c.segDepthLimit = depth
 		return nil
 	}
 }
@@ -118,7 +135,7 @@ func RateLimiter(limiterFor RateLimiterFor) Option {
 }
 
 // LatestSyncHandler defines how to store the latest synced cid for a given peer
-// and how to fetch it. Legs gaurantees this will not be called concurrently for
+// and how to fetch it. Legs guarantees this will not be called concurrently for
 // the same peer, but it may be called concurrently for different peers.
 type LatestSyncHandler interface {
 	SetLatestSync(peer peer.ID, cid cid.Cid)
@@ -153,6 +170,7 @@ type syncCfg struct {
 	alwaysUpdateLatest bool
 	scopedBlockHook    BlockHookFunc
 	rateLimiter        *rate.Limiter
+	segDepthLimit      int64
 }
 
 type SyncOption func(*syncCfg)
@@ -163,6 +181,13 @@ func AlwaysUpdateLatest() SyncOption {
 	}
 }
 
+// ScopedBlockHook is the equivalent of BlockHook option but only applied to a single sync.
+// If not specified, the Subscriber BlockHook option is used instead.
+// Specifying the ScopedBlockHook will override the Subscriber level BlockHook for the current
+// sync.
+// Note that calls to SegmentSyncActions from bloc hook will have no impact if segmented sync is
+// disabled.
+// See: BlockHook, SegmentDepthLimit, ScopedSegmentDepthLimit.
 func ScopedBlockHook(hook BlockHookFunc) SyncOption {
 	return func(sc *syncCfg) {
 		sc.scopedBlockHook = hook
@@ -172,5 +197,17 @@ func ScopedBlockHook(hook BlockHookFunc) SyncOption {
 func ScopedRateLimiter(l *rate.Limiter) SyncOption {
 	return func(sc *syncCfg) {
 		sc.rateLimiter = l
+	}
+}
+
+// ScopedSegmentDepthLimit is the equivalent of SegmentDepthLimit option but only applied to a
+// single sync.
+// If not specified, the Subscriber SegmentDepthLimit option is used instead.
+// Note that for segmented sync to function at least one of BlockHook or ScopedBlockHook must be
+// set.
+// See: SegmentDepthLimit.
+func ScopedSegmentDepthLimit(depth int64) SyncOption {
+	return func(sc *syncCfg) {
+		sc.segDepthLimit = depth
 	}
 }
--- a/selector.go
+++ b/selector.go
@@ -75,6 +75,102 @@ func getStopNode(selNode datamodel.Node) (datamodel.Link, bool) {
 	return stopNodeLink, err == nil
 }
 
+// getRecursionLimit gets the top-most recursion limit from the given selector node.
+// If no such selector is found, the returned bool will be false.
+func getRecursionLimit(selNode datamodel.Node) (selector.RecursionLimit, bool) {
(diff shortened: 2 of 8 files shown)


option.go
```diff
@@ -99,6 +102,27 @@ func BlockHook(blockHook BlockHookFunc) Option {
 	}
 }
 
+// SegmentBlockHook adds a hook that runs when a block is received along with a SegmentSyncActions.
+// Note that if segmented sync is disabled, calls on SegmentSyncActions will have no effect.
+// See: SegmentSyncActions, SegmentDepthLimit.
+func SegmentBlockHook(sbh SegmentBlockHookFunc) Option {
```
masih, line 108: The idea with this hook is to provide a way for the client to decode a synced CID and set the next segment sync CID from it. 

In general, I have taken a conservative approach of requiring explicit config over trying to auto-detect things since we don't know the shape of the general DAG being synced and there are a lot of ways things can go wrong.

option.go
```diff
@@ -174,3 +198,5 @@ func ScopedRateLimiter(l *rate.Limiter) SyncOption {
 		sc.rateLimiter = l
 	}
 }
+
+// TODO: add scoped variations of segmented sync options.
```
masih, line 202: Deferring this to a separate PR since I don't believe it's a blocker for resolving memory issues in storetheindex.

subscriber.go
```diff
@@ -811,11 +885,81 @@ func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wr
...
+	if h.subscriber.segDepthLimit < 0 && h.subscriber.segBlockHook != nil {
+		origLimit, limitOK = getRecursionLimit(sel)
+	}
+
+	// Revert back to sync without segmentation if original limit was not extracted, due to:
+	// - segment depth limit being negative; meaning segmentation is explicitly disabled, or
+	// - no segment block hook is configured; meaning we don't have a way to determin next
+	//   CID during segmented sync, or
+	// - the selector does not explore recursively, and has no top level recursion limit.
+	if !limitOK {
```
masih, line 904: The idea here is that segmented sync will only happen if it is enabled via config, i.e. both hook and depth for it are specified _and_ the original selector given to sync the dag with explores recursively with some recursion limit.

This is again on the conservative side to avoid accidental syncing of CIDs that the original selector might not have selected.

masih (commented):
*Self Review*

Marked the PR as draft while I'm working on its integration in storetheindex.

option.go
masih: Going to implement this since it is useful in storetheindex today.

option.go
gammazero, line 108: Would it be better to combine `SegmentBlockHook` and `SegemntDepthLimit` into one option since these need to be specified together?
```go
func SegmentBlockHook(sbh SegmentBlockHookFunc, depth int64) Option {
    ...
}
```

subscriber.go
```diff
@@ -778,15 +793,77 @@ func (h *handler) handleAsync(ctx context.Context, nextCid cid.Cid, syncer Synce
...
+
+	var syncedCids []cid.Cid
+	hook := func(p peer.ID, c cid.Cid) {
+		syncedCids = append(syncedCids, c)
+		if segbh != nil {
+			segbh(p, c, segSync)
+		}
+		if bh != nil {
+			bh(p, c)
+		}
```
gammazero, line 865: Does the calling order matter here?

subscriber.go
```diff
@@ -811,11 +888,77 @@ func (h *handler) handle(ctx context.Context, nextCid cid.Cid, sel ipld.Node, wr
...
+	var limitOK bool
+	var origLimit selector.RecursionLimit
+	// Only attempt to extract original recursion limit from selector if segmentation is enabled
+	// and there is a segment block hook.
+	//
+	// Not that we need the action in segment block hook to know which CID to sync in next segment.
+	// Therefore, it has to be set for segmented sync to function correctly.
+	if segdl > 0 && segbh != nil {
+		origLimit, limitOK = getRecursionLimit(sel)
+	}
```
gammazero, line 900: What if the caller wants to sync in one go (no segments), but also wants a recursion depth limit?  Does this mean that in order to have limited depth, a `SegmentBlockHookFunc` is required?

gammazero (commented):
Some questions to help me complete review

subscriber.go
masih: If the caller wants to sync in one go _and_ have a recursion depth the caller should use a selector with the desired recursion depth and disable segmentation.

Those two are separate things unless I am missing something?

option.go
masih: I wondered about doing this and settled on separate options. The rationale is:
* having separate options means they can be individually overridden, whereas if they are coupled the caller has to know both in order to override one.
* in the future if we come up with some default for the values that has segmentation enabled by default the options API would look more reasonable as separate options.
* it makes sense to separate them in the scoped variation, since either could be populated from a default set in `Subscriber` level, which then made the API look nicer (to me) if they are also split here.

Another option is to have both APIs. Happy to change it if this is a blocker.

subscriber.go
masih: Note that we are _not adding_ recursion here; we simply modify it to max segment depth. In fact no segmented sync will be performed at all if the original selector doesn't have some recursion limit regardless of whether segmentation is enabled or not.

subscriber.go
gammazero: Not sure why anyone would provide both functions, so was not sure if there was any implicit purpose for one or the other.

selector.go
```diff
@@ -75,6 +75,102 @@ func getStopNode(selNode datamodel.Node) (datamodel.Link, bool) {
...
+		}
+		return selector.RecursionLimitDepth(dl), true
+	}
+	if _, ok := err.(datamodel.ErrNotExists); ok {
+		_, err = selNode.LookupByString(selector.SelectorKey_LimitNone)
+		if err == nil {
+			return selector.RecursionLimitNone(), true
+		}
+	}
+	return selector.RecursionLimit{}, false
```
gammazero, line 106:
```suggestion
	if err != nil {
		if _, ok := err.(datamodel.ErrNotExists); ok {
			_, err = selNode.LookupByString(selector.SelectorKey_LimitNone)
			if err == nil {
				return selector.RecursionLimitNone(), true
			}
		}
		return selector.RecursionLimit{}, false
	}
	dl, err := dln.AsInt()
	if err != nil {
		return selector.RecursionLimit{}, false
	}
	return selector.RecursionLimitDepth(dl), true
}
```

gammazero (approved):
After much testing, this is working.  You can merge it now and address API issues in a separate PR or continue with the PR to handle API issues.

subscriber.go
```diff
@@ -427,11 +423,27 @@ func (s *Subscriber) OnSyncFinished() (<-chan SyncFinished, context.CancelFunc)
...
+	hook := cfg.scopedBlockHook
+	if s.generalBlockHook != nil && cfg.scopedBlockHook != nil {
+		cfg.scopedBlockHook = func(id peer.ID, c cid.Cid, a SegmentSyncActions) {
+			hook(id, c, a)
+			s.generalBlockHook(id, c, a)
+		}
+	} else {
+		cfg.scopedBlockHook = s.generalBlockHook
+	}
+
```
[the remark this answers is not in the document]
gammazero: We do not actually have a case where both are present, so there is no precedent.

I can imagine a scenario where someone may want to always have a general block hook called for any peer, but only sometimes wants to have a peer-specific function called. As you point out, `actions` can change the state of a sync, so there should not be two separate functions that can affect actions. So, I think there are two choices:
1. Only call one or the other function.
2. Call both functions but only pass `actions` to one.

Probably less surprising to have the scoped block hook override the general.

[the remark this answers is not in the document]
masih: Going with the override option. thanks

PR #130, opened 2022-05-25
Simplify rate limiter configuration

gammazero:
### Set rate limiter to use for sync only when creating Syncer
When doing a sync, set the rate limiter to the one provided for that individual sync, or if one was not provided, get it from the Subscriber's `rateLimiterFor` function if that function is available.

### Specifying a nil rate limiter is now OK
It is not necessary to construct an infinite rate limiter when there is no rate-limiting. A `nil` rate limiter avoids any rate-limiting logic.

### Stabilize tests by closing libp2p host connections
Tests were failing on some systems due to an error "socket: too many open files". This was caused by libp2p hosts having connections that were not closed. Tests have been modified to close the connections.

Also contains fix for #132

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -30,7 +30,7 @@ import (
 var log = logging.Logger("go-legs")
 
 // defaultAddrTTL is the default amount of time that addresses discovered from
-// pubsub messages will remain in the peerstore.  This is twice the default
+// pubsub messages will remain in the peerstore. This is twice the default
 // provider poll interval.
 const (
 	defaultAddrTTL       = 48 * time.Hour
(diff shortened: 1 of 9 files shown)


subscriber.go
```diff
@@ -196,14 +192,6 @@ func NewSubscriber(host host.Host, ds datastore.Batching, lsys ipld.LinkSystem,
...
 	scopedBlockHookMutex, scopedBlockHook, blockHook := wrapBlockHook()
 
-	if cfg.rateLimiterFor == nil {
-		// Allows all events. No rate limit.
-		limiter := rate.NewLimiter(rate.Inf, 0)
-		cfg.rateLimiterFor = func(publisher peer.ID) *rate.Limiter {
-			return limiter
-		}
-	}
-
```
gammazero, line 206: It is now OK to set the `rateLimiterFor` to `nil`.  in this case, any rate-limiting logic is skipped.

willscott (approved):

PR #131, opened 2022-05-26
Fix derived protocol ID so there are no double slashes

### LINKED ISSUE #86: head protocol ID ends up with double `/` for topic names starting with `/`

gammazero:
When the head client connects to a libp2p server and receives a "protocol not supported" error, it will try the old double-slashed protocol ID.

Fixes #86

### DIFF
--- a/p2p/protocol/head/head.go
+++ b/p2p/protocol/head/head.go
@@ -2,6 +2,7 @@ package head
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -10,13 +11,13 @@ import (
 	"sync"
 	"time"
 
-	logging "github.com/ipfs/go-log/v2"
-
 	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p-core/host"
 	peer "github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/protocol"
 	gostream "github.com/libp2p/go-libp2p-gostream"
+	multistream "github.com/multiformats/go-multistream"
 )
 
 const closeTimeout = 30 * time.Second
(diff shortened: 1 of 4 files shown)


p2p/protocol/head/head.go
```diff
@@ -63,7 +63,22 @@ func QueryRootCid(ctx context.Context, host host.Host, topic string, peerID peer
 				if err != nil {
 					return nil, err
 				}
-				return gostream.Dial(ctx, host, peerID, deriveProtocolID(topic))
+				conn, err := gostream.Dial(ctx, host, peerID, deriveProtocolID(topic))
+				if err != nil {
+					// If protocol ID is wrong, then try the old "double-slashed" protocol ID.
+					//
+					// TODO: remove this code when all providers have upgraded.
+					if err.Error() != "protocol not supported" {
```
masih, line 71: I suspect the error returned is [this](https://github.com/multiformats/go-multistream/blob/master/client.go#L18), which would make it possible to check the type instead of message.

gammazero: Thank you. I thought it might be [this one](https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/peer_loop.go#L21) which is not exported. After testing, looks like you are correct.

p2p/protocol/head/head_test.go
```diff
@@ -1,12 +1,11 @@
-package head_test
+package head
```
masih, line 1: It would be great to keep tests in a separate package where possible. It 
* helps maintain encapsulation, and
* showcases how to use the API in tests.

For a case where some internal API needs to be exposed for testing, like `deriveProtocolID`, there is already a nice pattern to expose just enough for testing purposes which I find to work really well. This is exposing APIs in a separate `*_test.go` file that is in the `head` package then using that exposed function in `head_test` package. [Example](https://github.com/filecoin-project/index-provider/blob/main/engine/chunker/cached_chunker_api_test.go). This way we achieve best of both words: maintained encapsulation between implementation and testing, and avoid exporting unwanted API just to be able to unit test.

masih (approved):

PR #133, opened 2022-05-26
Learn HTTP publisher address from announce message

### LINKED ISSUE #132: Subscriber `httppeerstore` is never updated with address from announcement messages
`Subscriber` keeps a separate peer store to store HTTP publisher addresses. This peer store [is only updated](https://github.com/filecoin-project/go-legs/blob/main/subscriber.go#L558) if at least one call to `Subscriber.Sync` is made with an explicit http multiaddr. Further, A publisher may change its address and announce it via the announce messages. 

This means even if a sync call with an explicit http multiaddr is made the address may change and is not updated until another call to sync with the updated http address.

Subscriber should therefore update the httppeerstore from announce messages.

masih:
Populate the http peer store with the address communicated via announce
messages when the announcer is indeed a HTTP publisher.

Otherwise, subscriber falls back on libp2p host
peerstore when no other address is given during an explicit sync call
and would end up using dtsync for a HTTP publisher.

Fixes #132

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -707,6 +707,10 @@ func (s *Subscriber) Announce(ctx context.Context, nextCid cid.Cid, peerID peer.
 
 	var syncer Syncer
 	if httpAddr != nil {
+		// Store this http address so that future calls to sync will work without a
+		// peerAddr (given that it happens within the TTL)
+		s.httpPeerstore.AddAddr(peerID, httpAddr, s.addrTTL)
+
 		syncer, err = s.httpSync.NewSyncer(peerID, httpAddr, limiter)
 		if err != nil {
 			return fmt.Errorf("cannot create http sync handler: %w", err)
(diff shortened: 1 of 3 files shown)


masih:
Captured [this](https://github.com/protocol/.github/issues/344) for build failure on MacOS caused during codecov coverage upload.

subscriber.go
```diff
@@ -708,6 +708,11 @@ func (s *Subscriber) Announce(ctx context.Context, nextCid cid.Cid, peerID peer.
 	var syncer Syncer
 	if httpAddr != nil {
 		syncer, err = s.httpSync.NewSyncer(peerID, httpAddr, limiter)
+
+		// Store this http address so that future calls to sync will work without a
+		// peerAddr (given that it happens within the TTL)
+		s.httpPeerstore.AddAddr(peerID, httpAddr, s.addrTTL)
```
gammazero, line 714: Seems like this should be moved to above the previous line.

PR #136, opened 2022-05-27
Bump version for rate limiter and test changes

gammazero:


willscott (approved):
feel free to update to v0.5 if you agree with goreleaser. this bump seems fine to me either way

gammazero:
I don't think anyone else is using the exported function that changed.

Issue #117, opened 2022-04-15
re-broadcast received http advertisements


willscott:
if a multisubscriber gets an HTTP ping from a publisher, it should re-package that advertisement into a gossip-sub message and send it onwards to others.
That will allow other subscriber instances to also learn about the update from the publisher

gammazero:
I assume that this means send the gossip pubsub message to the filecoin chain nodes for relay to other indexers (not directly to other indexers).  The problem with that is that the chain nodes expect any publisher to be verifiable on-chain, and indexers will not be.  So, this means there needs to be a way for chain nodes to validate pubsub messages coming from indexers.

willscott:
i think it's okay for the indexer to send it out over the topic as itself. Filecoin nodes may not validate / forward it, but if other indexers are peered directly to an indexer they could bootstrap from it. that seems sufficient, we don't don't need to go back through the filecoin gossip graph for this.

PR #139, opened 2022-06-17
Re-publish direct announce messages over pubsub

gammazero:
Add the ability to re-publish direct announce messages over pubsub. Direct announce messages are those that are sent directly to the API, and are not received over pubsub. Re-publishing these allows other indexers to receive the same announce messages that would otherwise only go to a single indexer.  Republishing can be optionally enabled/disabled using the ResendAnnounce option.

Fixes issue #117


subscriber.go
```diff
@@ -649,6 +688,24 @@ func (s *Subscriber) Announce(ctx context.Context, nextCid cid.Cid, peerID peer.
 	return nil
 }
 
+func (s *Subscriber) republish(ctx context.Context, nextCid cid.Cid, peerID peer.ID, addrs []multiaddr.Multiaddr) error {
+	pidStr := peerID.String()
+	buf := bytes.NewBuffer(nil)
+	buf.Grow(len("republish") + len(pidStr))
+	buf.WriteString("republish")
```
willscott, line 695: can we not add this to the ipld schema of the msg?

masih (commented):
A potential blocker I see is the reuse of field `ExtraData`, which was added as a way to facilitate gossipsub validation for messages published via `dtsync`.

Instead of repurposing that field, would it make sense to:
1. offer an option to `httpsync` Publishers to supply `ExtraData` just like we do in `dtsync` package.
    * So that on republish the gossip message can include it and help the message propagate through the networks that have validation enabled.
3.  dedicate a field in the message schema for the `republish` flag to make schema evolution easier.

gammazero:
@masih
> offer an option to httpsync Publishers to supply ExtraData just like we do in dtsync package.

That `ExtraData` would need to be a valid storage provider ID.  If they had this then they could publish directly over pubsub and would not need to send a direct announce to an indexer.

> So that on republish the gossip message can include it and help the message propagate through the networks that have validation enabled.

Validation checks that the sender's PeerID in the message matches the PeerID associated with the Lotus storage provider ID (SPID).  Republish would fail because the message would have the PeerID of the indexer republishing the message (this is not controlled by us) but the SPID of the original publisher.

masih:
All of that is true _for the current lotus network_ right?

I make no assumptions about the current usage. All I am saying is that from a generic library perspective, which go-legs strives to be, It seems odd that ExtraData setting is only available in gossipsub.

If HTTP announces end up on gossipsub via republish, would it make sense to expose the same options as we do for the dtsync in httpsync.

subscriber.go
gammazero: That would be best, but I need a way to tell the cbor codec that the field is optional so that messages from lotus, that do not have it, do not appear [invalid](https://github.com/filecoin-project/go-legs/blob/main/dtsync/cbor_gen.go#L91-L93) because they are missing a field.  Is there some tag like
```go
`cbor:",omitempty"`
```
?

dtsync/message.go
```diff
@@ -7,12 +7,15 @@ import (
...
 
+//go:generate go run gen/main.go
+
 var ErrBadEncoding = errors.New("invalid message encoding")
 
 type Message struct {
 	Cid       cid.Cid
 	Addrs     [][]byte
 	ExtraData []byte `json:",omitempty"`
+	Republish bool
```
gammazero, line 18: Would it be possible to do this instead?
```go
type Message struct {                                                                                                                                                                                 
    Cid       cid.Cid                                                                                                                                                                                 
    Addrs     [][]byte                                                                                                                                                                                
    ExtraData []byte                                                                                                                                                                                  
    OrigPeer  string                                                                                                                                                                                  
}                                                                                                                                                                                                     
```
If `OrigPeer` != "" then it is a republish.  This way the original peer ID does not need to be encoded into `ExtraData`.

masih: Please edit at will; the commit by me was meant to show how to generate cbor :)

dtsync/cbor_gen.go
```diff
@@ -88,7 +93,7 @@ func (t *Message) UnmarshalCBOR(r io.Reader) error {
 		return fmt.Errorf("cbor input should be of type array")
 	}
 
-	if extra != 3 {
+	if extra != 4 {
```
gammazero, line 96: This will be a problem with messages coming from non-updated lotus, since they will only have 3 fields.

subscriber.go
gammazero: I wrote the cbor_gen.go so that it is backward compatible with the old messages, and detects whether or not the new field is present.  We can leave this custom codec in place until lotus is using the new message format.

gammazero:
@masih 
> All of that is true for the current lotus network right?

Yes. The validator looks up miner info by miner ID (storage provider ID), and checks that the returned peerID matches the one from the pubsub message, [here](https://github.com/filecoin-project/lotus/blob/2c3dbb239a55590970066d61e68aaed734c6e299/chain/sub/incoming.go#L613-L615)

So, this whole PR is really only useful when indexers are bootstrapped to other indexers, and can relay pubsub messages to eachother.  Since indexers do not have storage provider IDs, they cannot publish pubsub messages to the filecoin mainnet nodes.

dtsync/cbor_gen.go
```diff
@@ -18,15 +18,22 @@ var _ = cid.Undef
...
+		var lengthBufMessage = []byte{131}
+		if _, err := w.Write(lengthBufMessage); err != nil {
+			return err
+		}
+	} else {
+		var lengthBufMessage = []byte{132}
+
+		if _, err := w.Write(lengthBufMessage); err != nil {
+			return err
+		}
```
masih, line 36: These are manual edits right? 

Is there any chance we can move these out of `_gen` file into a separate go file in the same package so that `go generate` can't override them?

If we can't then I recommend remocing `_gen` from the name and heavily document the rationale behind modifying generated code.

masih (commented):
I would love to avoid modifying generated code while keep using it 🤞 

Question:

Do you think it would work if we have: 
* two message structs with `cbor_gen` generated for each: one for the old struct and one for the new one, 
* a wrapper class that attempts to decode the message as the most common struct type which on `cbor input had wrong number of fields` error falls back onto the other.

That way we can continue to take advantage of generated code, i.e. don't need to maintain our own fork of it as cbor-gen library changes, while avoiding breaking wire protocol changes.

If that's not going to work then it would be great to rename the `cbor_gen` file to something that does not sound like it's generated and document the modifications heavily to avoid accidental override via `go generate` or `go run gen.go`.

Thank you for taking the time to make sure changes are backward compatible 🙇

masih (approved):
Approving based on Colo discussion

dtsync/cbor_gen.go
gammazero: I implemented the two message structs + wrapper, and afterward, decided that more code actually made it less clear to me. It also required copying the data to deserialize since deserialization consumes the buffer.

I opted to use non-generated code as this was less code to maintain, clearer (with comments added), and slightly more efficient.  Also, the message should at some point pre replaced with an IPLD schema and derialization done using bindnode.

Issue #144, opened 2022-08-05
A libp2p host cannot be a publisher and subscriber


masih:
Currently a single libp2p host can't be a legs publisher and subscriber. 

* when instantiated without an existing datatransfer, both publisher and subscriber instantiate their own where one overrides the other and results in odd behaviour at runtime. 
* and they cannot be instantiated from an existing datatransfer manager and graphsync exchange because on instantiation both publisher and subscriber try to register legs vouchers.

Ideally, the API should allow a single host to be both publisher and subscriber from an existing datatransfer manager instance without the functionality clash between the two.

PR #145, opened 2022-08-05
Handle already registered dtsync vouchers gracefully

masih:
It is possible for the same host to be a legs publisher and a lengs
subscriber. Instantiating both on the same host with an existing
datatransfer manager will attempt to register the vouchers twice which
results in a failure.

Match the body of the error returned by datatransfer and handle it
gracefully. Note that currently that seems to be the best we can do
since there are no APIs exposed by the manager to check if a type is
already registered or re-register a voucher type.

Fixes: https://github.com/filecoin-project/go-legs/issues/144


willscott (approved):

masih:
Note that the PR also bumps the version to get a concrete version needed in index-provider

willscott:
sounds good

PR #146, opened 2022-08-10
Option to filter un-routable IP addresses from announce messages

gammazero:
Option to filter private, loopback, and unspecified IP addresses from announce messages.

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -11,6 +11,7 @@ import (
 	"github.com/filecoin-project/go-legs/dtsync"
 	"github.com/filecoin-project/go-legs/gpubsub"
 	"github.com/filecoin-project/go-legs/httpsync"
+	"github.com/filecoin-project/go-legs/mautil"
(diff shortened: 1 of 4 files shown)


subscriber.go
```diff
@@ -694,6 +696,9 @@ func (s *Subscriber) watch(ctx context.Context) {
 			log.Infow("Handling pubsub announce", "peer", srcPeer)
 		}
 
+		if s.filterIPs {
```
willscott, line 699: shouldn't we also do this on the publisher?

gammazero: The addresses that come in on the announce message are the _publisher_ addresses.  The _provider_ addresses are supplied by the advertisement and the same needs to be done in storetheindex when updating the registery info with the addresses from the latest advertisement.  That's why I put the filter code in a separate package so sti could import it.

willscott (approved):

PR #155, opened 2022-09-10
add local check to dtsync

willscott:
graphsync syncer should check if requested dags are already fully available locally before re-transferring them.

### DIFF
--- a/dtsync/syncer.go
+++ b/dtsync/syncer.go
@@ -3,11 +3,17 @@ package dtsync
 import (
 	"context"
 	"fmt"
+	"io"
 	"time"
 
 	"github.com/filecoin-project/go-legs/p2p/protocol/head"
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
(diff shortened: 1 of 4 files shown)


ischasny (approved):
LGTM

masih:
Fixes https://github.com/filecoin-project/storetheindex/issues/691

willscott:
note that it mitigates the storetheindex issue by only downloading at the 5000-advertisement interval that synchronization happens at. that should probably be mostly enough mitigation, but it is still a different semantic from the http one.

dtsync/syncer.go
```diff
@@ -84,3 +98,45 @@ func (s *Syncer) Sync(ctx context.Context, nextCid cid.Cid, sel ipld.Node) error
...
+	// trusted because it'll be hashed/verified on the way into the link system when fetched.
+	getMissingLs.TrustedStorage = true
+	getMissingLs.StorageReadOpener = func(lc ipld.LinkContext, l ipld.Link) (io.Reader, error) {
+		r, err := s.ls.StorageReadOpener(lc, l)
+		if err == nil {
+			// Found block read opener, so return it.
+			return r, nil
+		}
+
+		return nil, fmt.Errorf("not available locally")
```
gammazero, line 114: nit: [Indent error flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow)
```suggestion
		if err != nil {
			return nil, fmt.Errorf("not available locally: %w", err)
		}
		// Found block read opener, so return it.
		return r, nil		
```

gammazero (approved):

PR #157, opened 2022-09-14
fix: respect context cancellation in dtsync Syncer.Sync()

elijaharita:
Context cancellation was not working for me with dtsync, I think this is because even though a context is passed into `OpenPullDataChannel()`, the function is not blocking and is only used for starting the channel, after which point the context no longer has any control over the channel.

The solution I found is to also watch for context cancellation next to syncDone, manually signal sync done if so, and then wait for its error again after.

This change fixed my issue in my project and context cancellation now works as expected.

### DIFF
--- a/dtsync/syncer.go
+++ b/dtsync/syncer.go
@@ -68,7 +68,7 @@ func (s *Syncer) Sync(ctx context.Context, nextCid cid.Cid, sel ipld.Node) error
 		select {
 		case err = <-syncDone:
 		case <-ctx.Done():
-			s.sync.signalSyncDone(inProgressSyncK, nil)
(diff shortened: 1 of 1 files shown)


gammazero (commented):
This is great!  I think this fixes https://github.com/filecoin-project/storetheindex/issues/738

Should `ctx.Err()` be sent to the `err` chanel?

elijaharita:
@gammazero param 2 in `signalSyncDone()`? I think you're right, if that's what you mean.

elijaharita:
@gammazero thanks just wanted to check, I applied that change

gammazero (approved):

PR #158, opened 2022-09-15
Add a test that checks CIDs match block content

masih:
Add a test that checks CIDs returned by nft.storage match hash of block content.


willscott (approved):

ischasny (approved):
LGTM

PR #169, opened 2022-11-04
Refactor announce handling into separate package

gammazero:
Refactors functionality for receiving, filtering, and resending announce messages. Logic moved to a new "announce" package with related sub-packages. The announce package lives in the storetheindex repo.

The announce package's `Receiver` object receives pubsub and direct announce messages, and delivers them all to the same channel from which they are retrieved using `Receiver.Next`.

The main reason for this refactor is to make announce message handling logic available to the new indexer coordination service. The package is located in storetheindex, and not in go-legs, to move consolidating shared indexing packages/libraries into storetheindex.

At lease one issue was also fixed: the subscriber would shutdown a pubsub topic that it did not own.


announce_test.go
```diff
@@ -113,12 +113,23 @@ func TestAnnounceReplace(t *testing.T) {
...
+	// announce handler goroutine delivered the announce to Subscriber for
+	// handling.
+	for i := 0; i < 10; i++ {
+		hnd.qlock.Lock()
+		pendingCid = hnd.pendingCid
+		hnd.qlock.Unlock()
+		if pendingCid == lastCid {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
```
willscott, line 128: what if you miss the pendingCid? is it possible for it to get over-written in a 10ms window?

gammazero: Fixed by exiting the loop only when the CID stops changing for long enough to be certain it is the last value that will be seen. Then that last value is checked against the expected.

dtsync/publisher.go
```diff
@@ -10,8 +10,8 @@ import (
 	"time"
 
 	dt "github.com/filecoin-project/go-data-transfer"
-	"github.com/filecoin-project/go-legs/gpubsub"
 	"github.com/filecoin-project/go-legs/p2p/protocol/head"
+	"github.com/filecoin-project/storetheindex/announce/gossiptopic"
```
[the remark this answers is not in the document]
gammazero: This PR fixes some defects as well as better organized the code, so I want go-legs to use the new announce handling code. There are a few options:
- Copy the `announce` package code into go-legs, so that it is present in both storetheindex and go-legs. Then port the other parts of go-legs into storetheindex and sever any dependency between the two.
- Put the `announce` package into some other repo that both go-legs and storetheindex can depend on.
- Move the `announce` package into go-legs and have storetheindex (coordination service) import the `announce` package from go-legs until.  ...and sometime in the future port go-legs into storetheindex.

masih (approved):

